PR #9, opened 2021-01-21
Test of golangci-lint

viccuad:


viccuad:
Closing, opened by mistake against here instead of my fork.

Issue #8, opened 2021-01-21
Add missing License and Copyright headers


dragonchaser:
Add license headers to all files/fix them up with the correct template.

dragonchaser:
@mattfarina we need some directions here, should we go with a standard apache license header or some suse specific thing?

kkaempf:
For license, we should choose a standard one (Apache, GPL, ...). For copyright, we need to use the standard SUSE copyright line.

kkaempf:
https://opensource.suse.com/suse-open-source-policy.html gives some directions, esp. about copyright:

> Copyright
> 
> The code you contribute upstream or publish as Open Source as part of your job at SUSE in your work time is copyrighted or
> otherwise owned by SUSE. Use “SUSE LLC” when attributing the copyright. Use your “@suse.com” email address for
> contributions.

PR #12, opened 2021-01-21
Add CLI with colors

### LINKED ISSUE #1: Setup CLI with color
Zypper has color output. This isn't unusual in modern CLI interfaces. The CLI will often try to detect if colors can be used and only use them when it will work. Along with this there is often an opt-out. Colors, such as red, can make it easier to spot errors.

This issue is to add that capability to `cmd/hypper`. A 3rd party library should be used for this.

dragonchaser:
This PRs adds a basic CLI skel with colors to hypper.

Closes #1


cmd/require/args.go
```diff
@@ -0,0 +1,88 @@
+/*
+Copyright The Helm Authors.
```
Itxaka, line 2: headers should be changed unless this is directly lifted from helm?

Itxaka (approved):

cmd/require/args.go
dragonchaser: Yes, we have an open issue about license headers #8 it is being looked into, will fix as soon this has been clarified.

Itxaka: :+1:

viccuad: I would drop it, considering that it all is boilerplate code and hardly copyrightable.

viccuad: I would even drop everything in `require/args*`, I don't think we are using it for now.

cmd/hypper/hypper.go
```diff
@@ -1,7 +1,42 @@
...
+	"os"
+
+	"github.com/fatih/color"
+	"github.com/mattfarina/hypper/pkg/cli"
+	"github.com/spf13/cobra"
+)
+
+var settings = cli.New()
+
+var red = color.New(color.FgRed).SprintFunc()
```
Itxaka, line 15: you can use `//nolint:golint,unused` here I think to silence the linter for now

Itxaka (approved):

PR #13, opened 2021-01-21
Adding design doc

mattfarina:


docs/design.md
```diff
@@ -0,0 +1,156 @@
...
+important things.
+
+### Optional Dependencies
+
+In Helm there are two ways to handle dependencies.
+
+1. Explicitly declare other charts as dependencies. This will cause those charts
+   to be installed. When the templates are rendered they happen in the same
+   sandboxed environment.
+2. Install charts separately and use values to connect them to each other. For
```
Itxaka, line 73: How do we plan to store this information across runs? Are we gonna use a CRD for hypper? local database? on-the-fly search?

How are we expecting this to work? Automatically? Will it require to put restrictions in place in case there is a existing X that we **dont** want other charts to use?

docs/design.md
```diff
@@ -0,0 +1,156 @@
...
+   example, install PostreSQL with one chart and then pass connection information
+   into another chart using a Secret or values that generate a Secret.
+
+Hypper intends to add a 3rd form of dependency via an optional dependency. For
```
Itxaka, line 77: Do we expect this to be used?

Like I can understand the reasoning behind it, but that means that you are not testing the same deployment on dev vs prod and that your values.yaml have to be different as well (to take care of disabling that chart deployment)

Is hypper expected to automatically disable that chart if the dependencies are optional and not installed or does that fall into the end user to do? How do we plan to account for the errors/issues that can come from that mismatch?

docs/design.md
```diff
@@ -0,0 +1,156 @@
...
+CRD chart or your chart can declare an optional dependency on a chart that
+installs Prometheus. When using hypper to install your chart you will be asked
+about installing the optional dependencies. In development you can choose not
+to install the dependencies while in production you can choose to install them.
+
+The optional dependency handling will extend to all commands.
+
+Optional dependencies will be specified using Helm chart annotations.
+
+### Shared Dependencies
```
Itxaka, line 89: isnt this overlapping with point 2 on optional dependencies? Sounds similar to me!

docs/design.md
```diff
@@ -0,0 +1,156 @@
...
+and custom environments you need a registry to store your images. If that can be
+used to store charts instead of a separate system you have fewer system
+dependencies.
+
+Hypper will make OCI registries a primary element in the system.
+
+One notable disadvantage to using an OCI registry for storing charts is a lack of
+an `index.yaml` file. This file has metadata enabling some functionality that
+can't be duplicated with an OCI registry due to APIs not existing (e.g., search).
+To restore these features we are looking to put an index into an OCI registry.
```
Itxaka, line 123: is it possible to expand on this? Does this mean pushing upstream to support having an index.yaml file or something completely different?

docs/design.md
```diff
@@ -0,0 +1,156 @@
...
+
+- JSON is faster to parse and uses less memory than YAML does. An `index.json`
+  file would faster and less memory intensive for Helm to work with than
+  `index.yaml`. A change like this could be done in a backwards compatible
+  manner.
+- `index.yaml` files are regularly pulled down to get updates and this is a
+  manual step. Using HTTP headers we can check if the content changed and only
+  pull it in those cases. By doing that some of the updates can be automated,
+  as well. A header that could be used, for example, is `etag`.
+- Charts can be stored locally in a cache and used from there rather than being
```
Itxaka, line 153: :+1:

docs/design.md
viccuad: > Is hypper expected to automatically disable that chart if the dependencies are optional and not installed or does that fall into the end user to do?

I side with a user option (reminds me of `Recommends` and `Suggests` package relations in .deb, that can be optionally installed if configured to do so).

> How do we plan to account for the errors/issues that can come from that mismatch?

We could solve for all the charts to be installed, before installing anything on the cluster, for their APIs inputs and outputs. And error if there can't be met.

docs/design.md
```diff
@@ -0,0 +1,156 @@
...
+Hypper can act as a testing ground for Helm. We do not expect that every feature
+will fit in upstream Helm.
+
+A second element in the difference is that Helm, as a package manager, does not
+handle environment deployment operations. Those are left to higher level tools
+like Helmfile (a push model like Ansible) and the Flux Helm Operator (a pull
+model like Chef). Helm expects higher level tools will use Helm and charts for
+operating in different environments.
+
+Hypper will move will make some assumptions that move it more towards the
```
Itxaka, line 28: do we expect to have conf/env files like Helmfile to declare the full environment and the relationships between charts?

docs/design.md
```diff
@@ -0,0 +1,156 @@
...
+
+Hypper intends to add a 3rd form of dependency via an optional dependency. For
+example, you are developing something that requires a prometheus CRD. You can
+depend on a chart that provides that CRD but does not provide prometheus. The
+CRD chart or your chart can declare an optional dependency on a chart that
+installs Prometheus. When using hypper to install your chart you will be asked
+about installing the optional dependencies. In development you can choose not
+to install the dependencies while in production you can choose to install them.
+
+The optional dependency handling will extend to all commands.
```
dragonchaser, line 85: How are we going to provide directives on which optional dependencies to be installed?
Interactive through the cli? A deployment manifest?

docs/design.md
mattfarina: This section of the document is talking about Helm rather than hypper. The design of this was all through through so I can talk to the details.

> How do we plan to store this information across runs? Are we gonna use a CRD for hypper? local database? on-the-fly search?

CRDs are an extension of the Kubernetes API. Extending the API is something that's usually only reserved for local dev and cluster admins. Most app operators don't have this kind of access and shouldn't. So, Helm does not assume this access.  We talked to people at companies about it.

Helm uses Secrets as the default with a configurable option to use PostgreSQL. A database is hard because you have to have that setup first. Where does it live? How is access handled? There is a fair amount of setup before you can get going. It's not like using an easy button.

With hypper we are working with the Helm library and Helm charts. I'm hoping with can piggy back on the existing storage setup Helm uses. That way we don't have to invent something new. I think we would look at a change when we encountered a need for it and then we'd have to think through the holistic experience.

If you're not familiar with how Helm does things I would suggest looking at it. I'm happy to answer any questions.

> How are we expecting this to work? Automatically? Will it require to put restrictions in place in case there is a existing X that we dont want other charts to use?

We can store metadata in Chart annotations. For example, have a default name and namespace to look for it in. All of the charts in a repo can be setup to look for each other. If the user installing has access to the namespace and resource we can connect them up. This would work with RBAC.

If you look at Rancher charts right now you'll see the use of annotations to store data. [For example, see this one](https://github.com/rancher/charts/blob/7a05185008739bd29da34f68b1d19027ba30287f/charts/rancher-logging/Chart.yaml).

docs/design.md
mattfarina: > > Is hypper expected to automatically disable that chart if the dependencies are optional and not installed or does that fall into the end user to do?
>
> I side with a user option (reminds me of Recommends and Suggests package relations in .deb, that can be optionally installed if configured to do so).

This is inspired by other package managers.

hypper and Helm are package managers rather than something like Helmfile, Ansible, Chef, or the Flux Helm Operator. it's a different concern.

Sometimes a local development environment can't run all the things. We are loading our clusters up with lots of services. On my local 16gb laptop I can't run everything. Sometimes I'll want a minimal install for dev. I would not consider my laptop running part of something to be the test that it's ready to go to prod. Someone should have another test.

Just like Helm, it's a tool and it's on users to make their own decisions.

I expect the install to be interactive and to disable that with flags that choose options. We will want smart defaults.

> How do we plan to account for the errors/issues that can come from that mismatch?

We prompt end users. Or, if in non-interactive mode we error out as early as we can. We want to be as safe as possible.

docs/design.md
mattfarina: Optional dependencies are ones that may not be installed. Shared dependencies are those that will be installed but they will be shared so we don't have multiple versions of something (e.g., istio) installed.

docs/design.md
mattfarina: > is it possible to expand on this? Does this mean pushing upstream to support having an index.yaml file or something completely different?

OCI registries, where you can push and pull containers from, implement the [distribution spec](https://github.com/opencontainers/distribution-spec). It has a catalog API that will list things you have access to. You need to page over it. Imagine doing that on docker hub? In practice it does not work. There is no search API and I don't expect one in the near future.

We are thinking of taking the `index.yaml` file for a Helm repo and putting it into an OCI registry like an image or Helm chart. Then you can add a Helm repo by telling Hypper to look at the OCI endpoint (e.g., `hypper repo add example oci://ghcr.io/mattfarina/examplerepo`).

Working with OCI registries this way is like working with object storage.

If you have some time take a look at `helm search`. We cannot do that with OCI, unfortunately.

When you have the metadata in an `index.yaml` file you can do a lot of resolution work without pulling down all the packages.

docs/design.md
mattfarina: We are not planning on it. Consider our current use cases...

1. Someone who wants to use a package manager. Do you want to maintain config files for zypper, homebrew, apt, or some other package manager? Do you ever touch config files for those?
2. As a library used in Rancher projects. There they will want to pass config around but not as files. And, Rancher is known for managing multiple clusters. We don't want to be in that business unless it's via fleet.

We want to keep the user experience in the realm of simple for end users. Whether they are using the API or CLI.

docs/design.md
mattfarina: > How are we going to provide directives on which optional dependencies to be installed? Interactive through the cli? A deployment manifest?

Annotations in Helm charts. Rancher already uses them as do others. This way it can be right there in the package.

docs/design.md
evrardjp: For naming, I suggest to also get inspired by package managers (for example python's pip), this would be clearer:
- Shared depencies -> System packages (to be opposed with "user" packages, which would here be better named as "namespaced packages", IMO)
-  Optional -> Recommends/Requires (This is more explicit).
WDYT?

docs/design.md
evrardjp: Can't the metadata be added into the image itself?
I think it would be good not rely on an index.yaml that's published into a OCI image. While the image brings the immutability and the nice layers, I think metadata of the possible images containing the helm chart is having a better isolation, and we won't have the issue "my index.yaml container is outdated". Everything would be just auto wired. WDYT?

docs/design.md
```diff
@@ -0,0 +1,156 @@
...
+to another for a package with some differences but still the same package.
+
+Helm does not make this form of assumption. It could be true. For example, you
+could install a chart from one location and then update from a fork in another
+location. Or two charts with the same name could be completely different. It
+is the responsibility of the end user to make this decision.
+
+Hypper is going to make the assumption that charts of the same name are the
+same chart from different locations.
+
```
evrardjp, line 138: This sounds like a terrible idea, without at least checking their signatures. 
If two helm charts have the same name and signed by the same authority, that makes kind of sense to assume they have been properly handled. Else it's opening the door to problems.

docs/design.md
```diff
@@ -0,0 +1,156 @@
...
+  `index.yaml`. A change like this could be done in a backwards compatible
+  manner.
+- `index.yaml` files are regularly pulled down to get updates and this is a
+  manual step. Using HTTP headers we can check if the content changed and only
+  pull it in those cases. By doing that some of the updates can be automated,
+  as well. A header that could be used, for example, is `etag`.
+- Charts can be stored locally in a cache and used from there rather than being
+  downloaded each time. This cache could be stored in places like a CI system
+  (e.g., like the way packages are stored for programming languages) or on a
+  local development computer.
```
evrardjp, line 156: What do you mean by this? Do you mean that hypper needs to understand Cache expiry/cache invalidation of OCI/webservers? Or do you mean that hypper needs to implement a "download (into cache)" function for air-gapped?

evrardjp (commented):
A few questions

docs/design.md
viccuad: The idea is to add an `index.yaml` as an object to the OCI registry, not to push an image with an `index.yaml` file to the registry.
Having an index in the registry means that one doesn't need to download and parse all the hypper objects of the registry to learn their dependency graph. This index.yaml object ideally would work for other object types in the OCI registry (container images, helm charts, etc), should they wish to use it. That would make it easier to upstream and have several OCI registry implementations understanding it.

docs/design.md
dragonchaser: I agree, as long you rely only on one catalogue this makes sense, but as soon you have external, maybe third-party deps, this is going to cause issues.

evrardjp: not more than if you were to upgrade a helm chart with the same name from a different location which does something completely different. "If you have different source, assume the helm charts are different" sounds fair to me.

docs/design.md
mattfarina: I like the idea of system packages and namespaced packages.

> Optional -> Recommends/Requires (This is more explicit).

Two thoughts come to mind...

1. We aren't building a new package manager. We are building on top of Helm and charts. So, we need to work around and with what we have today. The current `dependencies` list is equivalent to required ones.
2. The language around optional/recommends is something we can work out when we get to this feature. We should look at what others are doing in both the system and programming language package managers that are popular. From there we can make naming decisions around what will be easier for people to understand and figure out.

docs/design.md
mattfarina: > Can't the metadata be added into the image itself?

A few things come to mind...

1. How would a search feature work? In an OCI registry of any size it will be almost impossible to perform a search by looking at each item in the registry, looking at metadata, and creating a response. How would search work?
2. Having an `index.yaml` file means you can take the one file and calculate the dependency tree without having to look at the other locations. All of the metadata is there. If this is cached locally you can do this without going over the network. This could all be handled by going over the network. It would just take more time and you would require network access to do it.

The biggest reason for an index file is in order to be able to display a list of charts and to search it. If not for an index, how would you display a lot of things for an end user to see and decide to install? How would you perform search?

docs/design.md
mattfarina: > I don't understand the point 2: If the dependant package is only a helm library, there is no use of this, so there is indeed no need to pass file for the relationship, as the helm chart will declare itself the dependency.

Neither of these use cases are around managing the helm charts. They are both around how hypper itself is used. One case, the obvious one, is as a CLI that you use to install, update, and delete applications with.

The other use case... point 2... is when hypper itself is used as a library for other applications like rancher and fleet. For example, if fleet wanted to manage helm charts using hypper it would not call exec to a CLI. Instead it would import hypper as a software library and call Go APIs.

Does that make sense?

docs/design.md
mattfarina: We need to think about this in practice and what's good enough. If we looked for a perfect solution than this is not a good idea.

When it comes to names, this is already used in practice. I'll give both a Helm and non-Helm example...

1. When Helm deprecated the stable charts repository people needed to move to a new location for their charts. They were able to add a new repository and just upgrade from the new one instead. Checking the API and features was left as an exercise to the end user rather than being automated. But, it worked. There are also many companies that will install a chart from a company and then later fork or vendor it. Upgrading from their fork instead.
2. In zypper you have the ability to install something and then change vendors later. The package you change to can have functional differences. For example, I change my desktop codec packages to those that support proprietary codecs so I can browse the web and use A/V. Same package name but different vendors shipping differences.

Let's think of this in terms of use cases...

* As an operator at a company, I have forked a chart we are using for our internal use. We are already using it. We want to tell the system to switch to the one we have.
* As someone who added a feature to a chart that's not accepted upstream, I want to share my new location for the chart and give people an easy way to switch to my version.
* As someone consuming a chart, the organization I get it from has stopped updating it and I want to get updates from a new location. This could be because there are security issues or other problems. But, if the organization that owns it repairs the issues I would like to switch back.

How would a system work that accommodates these types of use cases?

docs/design.md
mattfarina: Hypper can have a local cache directory where it stores charts. If you already have the chart at a version in the cache than you can install it right from there rather than downloading it. The index has a digest (sha256) so we can likely make this a content addressable system. This could be something to make work better in upstream Helm.

docs/design.md
Itxaka: :+1:

docs/design.md
Itxaka: :+1:

PR #14, opened 2021-01-22
Emojis :pizza:

### LINKED ISSUE #2: Setup CLI with icons
minikube has icons. This isn't unusual in modern CLI interfaces. The CLI will often try to detect if icons can be used and only use them when it will work. For example, they should not be used in CI so the logs can be clear. Along with this there is often an opt-out. 

This issue is to add that capability to `cmd/hypper`. A 3rd party library should be used for this if possible.

dragonchaser:
This PR adds support for outputting emojis to the cli.
It uses the common `:emoji:` notation known from popular messaging clients like element.

examples:
`:+1: `-> :+1: 
`:pizza:` -> :pizza: 
`:100:` -> :100: ....

Closes #2

### DIFF
--- a/cmd/hypper/hypper.go
+++ b/cmd/hypper/hypper.go
@@ -27,7 +27,7 @@ func main() {
 	}
 
 	if err != nil {
-		logger.Debug(eycandy.Magenta("%v"), err)
+		logger.Debug(eyecandy.Magenta("%v"), err)
 		os.Exit(1)
 	}
 
@@ -42,7 +42,7 @@ func main() {
 	})
 
 	if err := cmd.Execute(); err != nil {
-		logger.Debug(eycandy.Magenta("%v"), err)
+		logger.Debug(eyecandy.Magenta("%v"), err)
 		os.Exit(1)
 	}
 }
--- a/pkg/eyecandy/emoji.go
+++ b/pkg/eyecandy/emoji.go
@@ -1,0 +1,26 @@
+package eyecandy
+
+import (
+	"fmt"
+	"github.com/kyokomi/emoji/v2"
+	"regexp"
+)
+
+func ESPrintf(emojisDisabled bool, format string, v ...interface{}) string {
+	if emojisDisabled {
+		return fmt.Sprintf(removeEmojiFromString(format), v...)
+	}
+	return emoji.Sprintf(format, v...)
+}
+
+func ESPrint(emojisDisabled bool, s string) string {
+	if emojisDisabled {
+		return fmt.Sprint(removeEmojiFromString(s))
+	}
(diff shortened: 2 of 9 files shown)


pkg/eyecandy/emoji.go
```diff
@@ -0,0 +1,26 @@
+package eyecandy
+
+import (
+	"fmt"
+	"github.com/kyokomi/emoji/v2"
+	"regexp"
+)
+
+func ESPrintf(emojisDisabled bool, format string, v ...interface{}) string {
+	if emojisDisabled {
```
Itxaka, line 10: This kind of sucks, having to pass this flag all the time. Isnt there a way we can directly obtain it here from the settings or environment?

viccuad: Agree here.

dragonchaser: I know, I found 2 ways, either moving the settings into global variable, which sucks in other ways or adding the ESPrint functions to the environment.go, where they clearly do not belong. So I opted for the least sucking way :)

pkg/eyecandy/emoji.go
```diff
@@ -0,0 +1,26 @@
...
+
+func ESPrint(emojisDisabled bool, s string) string {
+	if emojisDisabled {
+		return fmt.Sprint(removeEmojiFromString(s))
+	}
+	return emoji.Sprint(s)
+}
+
+func removeEmojiFromString(s string) string {
+	re := regexp.MustCompile(`:[a-zA-Z0-9-_+]+?:`)
```
viccuad, line 24: Not a showstopper for now (and I'm half guilty of this anyways :sweat_smile:), but I wonder of the speed cost of the lookahead on the regexp, every time we call it on logs.

pkg/eyecandy/emoji.go
dragonchaser: So how do we resolve this? Keep it this way for now & merge?

cmd/hypper/hypper.go
```diff
@@ -27,7 +27,7 @@ func main() {
 	}
 
 	if err != nil {
-		logger.Debug(eycandy.Magenta("%v"), err)
```
Itxaka, line 30: wait, this was on main and it didnt fail the ci?? How is that possible with the strictness of go??

Itxaka (approved):
Good to me!

cmd/hypper/hypper.go
dragonchaser: No because the package itself contained the typo in the packagename, so it was wrong but still correct code, fixed the packagename aswell.

PR #15, opened 2021-01-22
Enable matrix for go version testing

Itxaka:
We should test at least the maintained go versions so we can check
if something fails on any of those versions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go-version: [1.15.x]
+        go-version: ["1.16.0-beta1", "1.15.7", "1.14.14"]
         os: [macos-latest, windows-latest, ubuntu-latest]
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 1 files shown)


Itxaka:
This wont work until there is tests of course.

Itxaka:
nice, we can even test betas :o

.github/workflows/ci.yml
```diff
@@ -10,7 +10,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go-version: [1.15.x]
+        go-version: [1.16.0-beta1, 1.15.7, 1.14.14]
```
viccuad, line 13: We decided to use 1.15.7 as oldest, see https://github.com/mattfarina/hypper/pull/11#discussion_r561987772.

viccuad (commented):
Yay! first PRs. Welcome :)

.github/workflows/ci.yml
Itxaka: That seems to be 1.15 vs 1.13 but supported versions for go are release and release minus 1, so Im guessing we want to have that as well? Like we want to make sure we cover both stable releases ,plus Im sure down the line the SLES supported go version its on the requirements and that sometimes lag a bit compared to the latest go version :)

Itxaka: just to have it written down, the reason for using 1.15.X only was a security bug vs 1.13.X

That bug is fixed in 1.14.14 so it should be safe to use.

viccuad: Sounds good, resolving the conversation :).

Itxaka:
now it works even with the beta due to the coverage run not failing if there are no tests :D

viccuad (approved):
Sounds good! merging :)

PR #17, opened 2021-01-22
Drop coverage script and make it simpler

### LINKED ISSUE #16: Clean coverage script
Currently the coverage.sh is lifted from the helm package if Im not mistaken. It includes a lot of extra stuff that we are currently not using.

We should probably remove the scripts part as its not needed currently and make it check the cover in the Makefile directly. 

Then we can iterate over what to use and whatnot in the future from a clean slate.

Itxaka:
We currently dont need a separate script to run the coverage
as we arent even tracking it.

This PR drops the uneeded script and makes the cover action more
clear and direct to the point

Partially Fixes #16


viccuad (approved):
Yep, totally right :).

mattfarina (approved):

PR #21, opened 2021-01-27
Add install command (without tests)

viccuad:
This PR adds the install subcommand, wrapping Helm calls to install a Helm chart.
It does that by including a helmSettings pointer in Hypper's EnvSettings struct. In the future, we may have an opinionated helmSettings constructor.

Tests in `install_test.go` are missing still. Tested manually by installing a helm chart against a cluster with `hypper install nginx suse/nginx-ingress`.


cmd/hypper/install.go
```diff
@@ -0,0 +1,131 @@
...
+	return cmd
+}
+
+func runInstall(args []string, client *helmAction.Install, valueOpts *values.Options, out io.Writer) (*release.Release, error) {
+	var helmSettings = helmCli.New()
+	fmt.Println(client.Version)
+
+	// TODO add hypper specific code here
+
+	if client.Version == "" && client.Devel {
```
Itxaka, line 51: Sorry I dont understand this. Is this part of helm upstream or are we using it to skip over charts not having versions?

cmd/hypper/testdata/empty/Chart.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
```
Itxaka, line 1: Im guessing this is a example chart to use for testing down the line no?

cmd/hypper/install.go
viccuad: This comes from Helm's https://github.com/helm/helm/blob/master/cmd/helm/install.go#L174-L177.
Seems to be a catch-all in case that a chart version was not specified up until here.

cmd/hypper/testdata/empty/Chart.yaml
viccuad: Correct, could have been left out of the PR.

cmd/hypper/install.go
Itxaka: mmmh, I was looking at the wrong part `pkg/action` :facepalm: 

Thanks!

Itxaka (approved):
looking good!

viccuad:
merging then!

viccuad:
Partially implements https://github.com/mattfarina/hypper/issues/3, besides tests.

PR #22, opened 2021-01-27
Add install tests

Itxaka:
Currently because we dont have any custom code, this is mainly
a 1:1 copy from upstream helm tests, but its still useful to
make sure that we are using the helm libs properly and not
breaking stuff when we start to manipulate charts and so on


viccuad:
Closes https://github.com/mattfarina/hypper/issues/3

viccuad (approved):

PR #25, opened 2021-01-28
Enable verbose output

Itxaka:
Enables verbose so we can analyze any failures


viccuad (approved):
Approving, let's see if we are lucky and we see the behaviour once this is in.

PR #34, opened 2021-02-09
Updating golangci-lint to 1.36

mattfarina:
Note, I noticed the lint process was taking more than 29s and timing out. golangci-lint defaults to a 1m0s timeout but that was not happening. Helm takes about 45s to lint in CircleCI. So, I increased the lint timeout period.


viccuad (approved):

viccuad:
Seems that this doesn't solve it, as when this hit master, we got the errors again: https://github.com/mattfarina/hypper/runs/1869902485.

PR #39, opened 2021-02-12
moving to new log-go home

mattfarina:
The github.com/mattfarina/log-go package moved to its new home
in an organization. This change uses it in its new home.


dragonchaser (approved):
lgtm

PR #40, opened 2021-02-15
add file extension for windows

### LINKED ISSUE #36: Fileextension for windows executable
Windows needs a `.exe` extension to execute files. We need to adapt our `Makefile` to move the output `bin/hypper` file to `bin/hypper.exe` when run on windows.

dragonchaser:
This PR closes #36
It outputs `bin/hypper.exe` instead of `bin/hypper` when running make on windows.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ build: lint $(BINDIR)/$(BINNAME)
 SRC := $(shell find . -type f -name '*.go' -print) go.mod go.sum
 
 $(BINDIR)/$(BINNAME): $(SRC)
-	go build $(GOFLAGS) -tags '$(TAGS)' -ldflags '$(LDFLAGS)' -o '$(BINDIR)'/$(BINNAME)${GOEXE} ./cmd/hypper
+	go build $(GOFLAGS) -tags '$(TAGS)' -ldflags '$(LDFLAGS)' -o '$(BINDIR)'/$(BINNAME) ./cmd/hypper
 
 .PHONY: install
 install: build
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,9 @@
 BINDIR	:= $(CURDIR)/bin
-BINNAME	?= hypper
+ifeq ($(OS),Windows_NT)
+	BINNAME	?= "hypper.exe"
+else
+	BINNAME	?= "hypper"
+endif
 INSTALL_PATH ?= /usr/local/bin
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -36,7 +40,7 @@ build: lint $(BINDIR)/$(BINNAME)
 SRC := $(shell find . -type f -name '*.go' -print) go.mod go.sum
 
 $(BINDIR)/$(BINNAME): $(SRC)
-	GO111MODULE=on go build $(GOFLAGS) -tags '$(TAGS)' -ldflags '$(LDFLAGS)' -o '$(BINDIR)'/$(BINNAME) ./cmd/hypper
+	go build $(GOFLAGS) -tags '$(TAGS)' -ldflags '$(LDFLAGS)' -o '$(BINDIR)'/$(BINNAME)${GOEXE} ./cmd/hypper
```
viccuad, line 43:
```suggestion
	go build $(GOFLAGS) -tags '$(TAGS)' -ldflags '$(LDFLAGS)' -o '$(BINDIR)'/$(BINNAME) ./cmd/hypper
```
Seems that `${GOEXE}` is a leftover, `$(BINNAME)` will already contain the `.exe` at this point. If not, we should use `${GOEXE}` in the clean target too.

Makefile
```diff
@@ -36,7 +40,7 @@ build: lint $(BINDIR)/$(BINNAME)
 SRC := $(shell find . -type f -name '*.go' -print) go.mod go.sum
 
 $(BINDIR)/$(BINNAME): $(SRC)
-	GO111MODULE=on go build $(GOFLAGS) -tags '$(TAGS)' -ldflags '$(LDFLAGS)' -o '$(BINDIR)'/$(BINNAME) ./cmd/hypper
```
viccuad, line 39: I suppose removing `GO111MODULE=on` is ok :thinking: . Afterall, we are only building and testing on 1.14, 1.15, and it is default since 1.13.

viccuad (changes requested):

Makefile
dragonchaser: I have removed it, make on windows is a bit quirky and can not access the variables from `go env` for some reason. This was a leftover from an experiment.

Makefile
dragonchaser: Yes, also @mattfarina mentioned that we do not need it any more.

viccuad (approved):

PR #46, opened 2021-02-18
Use Golang composition to extend helm/pkg/action in hypper/pkg/action

viccuad:
Provide a manageable API to __hypper/pkg/action__ by embedding __helm/pkg/action__ types inside of it. Use Golang composition and embedded fields for doing so.

This allows us to expand the action package slowly from Helm's, while presenting a Hypper API that can be consumed externally.

See commit comments for expanded explanations.

Upsides:
- The refactor allows us to keep the same structure as Helm (for easy upstreaming).
- Allows us to extend the code incrementally.
- Allows us to provide the __hypper/pkg/*__ for consumption.

Downsides:
- We need to disable checking for composites in `go vet`. Given that we don't compose any other types, and the types that we compose (__hypper/pkg/action__ and __hypper/pkg/*__ in the future) are under our control, I believe the trade-off really pays off.

Alternatives considered:
- Don't use composite types, and wrap Helm methods and functions instead: IMHO this becomes unreasonable, as changing functions keep pulling other functions to be changed. Plus, makes upstreaming more difficult. Still, it would provide a consumable API for __hypper/pkg/*__.
- Keep consuming __helm/pkg/*__ directly for now. Simplifies current development. but this will pull a lot of copy-pasted Helm code, and doesn't create a consumable API for _hypper/pkg*_. Feels more like working on a Helm fork, than on a separate program that calls Helm libraries.

### DIFF
--- a/cmd/hypper/hypper.go
+++ b/cmd/hypper/hypper.go
@@ -12,6 +12,7 @@ import (
 	"github.com/rancher-sandbox/hypper/pkg/eyecandy"
 	"github.com/spf13/cobra"
 	"gopkg.in/yaml.v2"
+	helmAction "helm.sh/helm/v3/pkg/action"
 	kubefake "helm.sh/helm/v3/pkg/kube/fake"
 	"helm.sh/helm/v3/pkg/release"
 	"helm.sh/helm/v3/pkg/storage/driver"
@@ -23,7 +24,11 @@ func main() {
 	logger := logcli.NewStandard()
 	log.Current = logger
 
-	actionConfig := new(action.Configuration)
+	helmActionConfig := new(helmAction.Configuration)
+	actionConfig := &action.Configuration{
+		Configuration: helmActionConfig,
+	}
+
 	cmd, err := newRootCmd(actionConfig, log.Current, os.Args[1:])
 	if settings.Debug {
 		logger.Level = log.DebugLevel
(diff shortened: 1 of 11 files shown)


viccuad:
(I recommend that this PR is reviewed by stepping over each commit).

viccuad:
Maybe adding some lines in _docs/design.md_ with the upsides and downsides of using composition and deactivating the `-composite` lints would be a good idea.

cmd/hypper/hypper.go
```diff
@@ -23,7 +23,7 @@ func main() {
 	logger := logcli.NewStandard()
 	log.Current = logger
 
-	actionConfig := new(helmAction.Configuration)
+	actionConfig := new(action.Configuration)
```
viccuad, line 26: Unit tests pass, but running `hypper list` fails with an invalid mem address or nil pointer. I suppose we need to initialize our `hypper/pkg/action.Configuration` correctly here. Still, I'm looking at how to.

viccuad: fixed in https://github.com/rancher-sandbox/hypper/pull/46/commits/c7fd3ac77bdd38273cf9dec3ff1c26b6ad9e5305.

cmd/hypper/hypper.go
```diff
@@ -23,7 +24,11 @@ func main() {
 	logger := logcli.NewStandard()
 	log.Current = logger
 
-	actionConfig := new(helmAction.Configuration)
+	helmActionConfig := new(helmAction.Configuration)
+	actionConfig := &action.Configuration{
+		helmActionConfig,
+	}
```
mattfarina, line 30: If you change the way the embedding is handled you leave vet unchanged.

```go
actionConfig := &action.Configuration{
	Configuration: helmActionConfig,
}
```

There's a blog post with a nice example at https://dave.cheney.net/2015/05/22/struct-composition-with-go

This idea applies in the test, too. Then you can strip out the `go vet` modifications.

mattfarina (changes requested):

cmd/hypper/hypper.go
viccuad: doh, thanks :).

Done. Force-pushed to amend commit messages and drop some commits too.

mattfarina (approved):

PR #47, opened 2021-02-19
Add go 1.16.x to CI

viccuad:
I would like to add `latest` or `1.x` instead, but the GH Action doesn't support it yet, alas.


viccuad:
Well, I did the commit through GH ui, which sets `Author:     Victor Cuadrado Juan <viccuad@users.noreply.github.com>` and of course makes the DCO bot fail.

Force-pushed with correct author.

dragonchaser (approved):
lgtm

Issue #29, opened 2021-01-29
Enable charts to provide default location configuration


mattfarina:
Consider the case of installing a system wide prometheus chart and wanting it to be in a well known location. A Helm repository owner can craft a set of charts with default configuration that can be overridden for name and namespace. These would be specified using annotations in the chart:

```yaml
annotations:
  catalog.cattle.io/namespace: fleet-system
  catalog.cattle.io/release-name: fleet
```

This example is pulled from Ranchers own charts.

`hypper install` would need to take the name as a flag or optional argument where the flag or argument would override the default specified in the chart.

mattfarina:
After out call today with Darren I want to add the detail of:

On install use the annotations to get the name (a.k.a release name) and namespace to use for installation.

There are hypper annotations to use for these along with Rancher catalog fallback annotations. The hypper ones are:

```yaml
annotations:
  hypper.cattle.io/namespace: fleet-system
  hypper.cattle.io/release-name: fleet
```

The catalog fallback ones are:

```yaml
annotations:
  catalog.cattle.io/namespace: fleet-system
  catalog.cattle.io/release-name: fleet
```

The priority order for namespace should be:

1. The flag passed in from the CLI
2. The `hypper.cattle.io/namespace` annotation
3. The `catalog.cattle.io/namespace` annotation
4. The current namespace as configured with the kubeconfig (when no information is given). This is what Helm does when no namespace is passed in by the flag.

In Helm there is an argument for the name (unless `--generate-name` is set). Hypper will change from this.

The priority order for name should be:

1. The argument passed in from the CLI
2. The `hypper.cattle.io/release-name` annotation
3. The `catalog.cattle.io/release-name` annotation
4. `--generate-name`
5. An error as Helm does today

The name and namespace used should be set to output at an info level. We want this see all of the time.

For dev/test purposes you can use the current rancher catalog charts. The repository is at https://charts.rancher.io. It uses the annotation prefix of `catalog.cattle.io`.

PR #49, opened 2021-02-22
Add `hypper install` with support for annotations

viccuad:
- Implement `hypper install` with name, `--generate-name` and `--create-namespace` flags, Hypper and Helm annotations.
- Add minimal scaffolding for pkg/action/* and pkg/action/install_test.go to be able to implement `pkg/action/install*`.

Closes https://github.com/rancher-sandbox/hypper/issues/29.


viccuad:
Rebased, fixing _go.mod_ conflicts with main.
`golangci-lint` fails for things in _internal/registry_, which need to be fixed and upstreamed.

pkg/action/install.go
```diff
@@ -149,3 +149,11 @@ func (i *Install) Chart(args []string) (string, error) {
 	// len(args) == 1
 	return args[0], nil
 }
+
+// NameAndChart overloads Helm's NameAndChart. It always fails.
+//
+// On Hypper, we need to read the chart annotations to know the correct release name.
```
viccuad, line 155: @mattfarina (et al): is this needed?  I need to "overload" this function as it its usage with Hypper libs is incorrect. Since there doesn't seem to be any other way in Golang, "it eez what it eez".

mattfarina (changes requested):
I've not had a chance to do a full review but I tested the change and it's not working.

Using Helm I pulled the fleet charts from charts.rancher.io. I added the repo and pulled the chart locally.

Then I used the command `hypper install fleet-0.3.300.tgz` to install the chart. Instead of installing into the fleet namespace it installed into the default namespace. This did not matter if the namespace existed before running the command or not.

To reproduce:

```sh
$ helm repo add rancher-charts https://charts.rancher.io/
$ helm pull rancher-charts/fleet
$ hypper install fleet-0.3.300.tgz
```

The hypper install command gave the output:

```sh
Installing chart "fleet" in namespace "fleet-system"…
Done! 👏
```

Yet, it installed into default.

cmd/hypper/install.go
```diff
@@ -18,18 +20,32 @@ import (
 	"github.com/rancher-sandbox/hypper/pkg/eyecandy"
 )
 
-var installDesc = `install a helm chart by wrapping helm calls (for now)`
+const installDesc = `
+This command installs a chart archive.
```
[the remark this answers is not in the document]
viccuad: True. Fixed that and number typos on https://github.com/rancher-sandbox/hypper/pull/49/commits/2b5d5ee8a9b5eecc28520a39096d2fb4791cc021.

Itxaka:
> I've not had a chance to do a full review but I tested the change and it's not working.
> 
> Using Helm I pulled the fleet charts from charts.rancher.io. I added the repo and pulled the chart locally.
> 
> Then I used the command `hypper install fleet-0.3.300.tgz` to install the chart. Instead of installing into the fleet namespace it installed into the default namespace. This did not matter if the namespace existed before running the command or not.
> 
> To reproduce:
> 
> ```shell
> $ helm repo add rancher-charts https://charts.rancher.io/
> $ helm pull rancher-charts/fleet
> $ hypper install fleet-0.3.300.tgz
> ```
> 
> The hypper install command gave the output:
> 
> ```shell
> Installing chart "fleet" in namespace "fleet-system"…
> Done! 👏
> ```
> 
> Yet, it installed into default.

Looks like an opportunity to introduce end2end tests here to test that indeed our install is intalling to the expected locations for example :D

Itxaka:
woops, wrong button!

viccuad:
> Looks like an opportunity to introduce end2end tests here to test that indeed our install is intalling to the expected locations for example :D

Good point! Helm seems to have https://github.com/helm/acceptance-testing, but they don't seem to be run by their CI. We could open our own issue on adding e2e tests for book keeping.

But to be fair, the errors on this PR shouldn't have slipped, and that means that the test output is badly crafted. Fixing those would make the bug stand out :).

viccuad:
I can't reproduce `hypper install fleet-0.3.300.tgz`, on a fresh k3d cluster. Here it installs in the correct `fleet-system` namespace.
I have, though, found a different incorrect test, see https://github.com/rancher-sandbox/hypper/pull/49/commits/1ec1f482ea16953a9c2ce74f01af7ec4c1fbfaa3.

We will work in fixing that, which means correctly using `cli.settings` in `cmd/hypper/install.go`, and refactoring then `pkg/cli`.

pkg/action/install.go
```diff
@@ -23,21 +44,116 @@ func NewInstall(cfg *Configuration) *Install {
...
+}
+
+// SetNamespace sets the Namespace that should be used in action.Install
+//
+// This will read the chart annotations. If no annotations, it leave the existing ns in the action.
+func (i *Install) SetNamespace(chart *chart.Chart, defaultns string) {
+	i.Namespace = defaultns
+	if chart.Metadata.Annotations != nil {
+		if val, ok := chart.Metadata.Annotations["hypper.cattle.io/namespace"]; ok {
+			i.Namespace = val
```
mattfarina, line 69: This is one of the most confusing parts of the Helm API but this is not setting the namespace the kube client will put the object in. In `action.New` the namespace is set and this is the namespace that will end up being used to put the resources in. This is because `action.Init` creates a kubernetes client with a namespace set by a flag (in helm) or environment variable.

mattfarina (changes requested):
ok, you stumbled on a complicated part of the helm api.

pkg/action/install.go
viccuad: Thanks, I see now! (debugger ftw). The complication I have is that I have no way of knowing if the namespace already set up in the `action` comes from the flag (first priority case) or comes from the kubeconfig context (last priority case). The namespace logic was always out of _action/install.go_, but now with annotations (which is in the middle of priority cases), we ought to provide a getter at least. So I suppose the logic between picking from the `action.Namespace` and the annotations may be out of the _action/install.go_ library.

viccuad: We fixed it in the last 4 comments.

Itxaka (approved):
looks good to me!

pkg/cli/environment.go
```diff
@@ -7,65 +23,41 @@ import (
...
 
 // defaultMaxHistory sets the maximum number of releases to 0: unlimited
 const defaultMaxHistory = 10
 
-// EnvSettings describes all of the environment settings.
+// EnvSettings is a composite type of helm.pkg.env.EnvSettings.
+// describes all of the environment settings.
 type EnvSettings struct {
+
+	// we use Helm's for all exported fields
```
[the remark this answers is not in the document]
viccuad: To be fair I'm not sure we need a composite type here! All the fields need to be changed to use `HYPPER_` env vars, so we ended up reverting the commit that brings this.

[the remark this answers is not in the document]
viccuad: Reverted in https://github.com/rancher-sandbox/hypper/pull/49/commits/36ef5271ca84c78793885ca3d92c53d3ec903c6c/

mattfarina (dismissed):
I'm still unable to install fleet into the correct namespace. You can get the chart I'm testing with [here](https://github.com/rancher/charts/raw/dev-v2.5/released/assets/fleet/fleet-0.3.300.tgz).

When I try to use hypper install on this chart it installs into the default (i.e. currently selected) namespace instead of the one specified in the _Chart.yaml_ file.

Can we get this working?

Itxaka:
This is weird because I tested and it seemed to work:

```
# itxaka @ zeus in ~/projects/hypper on git:viccuad/add-annotations o [12:24:00] 
$ bin/hypper install fleet https://github.com/rancher/charts/raw/dev-v2.5/released/assets/fleet/fleet-0.3.300.tgz
Installing chart "fleet" in namespace "fleet-system"…
Done! 👏 

# itxaka @ zeus in ~/projects/hypper on git:viccuad/add-annotations o [12:24:14] 
$ bin/hypper list
NAME 	NAMESPACE   	REVISION	UPDATED                                	STATUS  	CHART        	APP VERSION
fleet	fleet-system	1       	2021-03-02 12:24:14.476796873 +0100 CET	deployed	fleet-0.3.300	0.3.3      
test 	default     	2       	2021-02-26 11:10:20.787363474 +0100 CET	deployed	empty-0.2.0  	           
test2	default     	1       	2021-02-26 11:08:09.799608082 +0100 CET	deployed	empty-0.1.0  	    ```

**Comment by Itxaka** [comment]
ok wow, no, this doesnt work no matter what hypper (or helm!) says!

```
# itxaka @ zeus in ~/projects/hypper on git:viccuad/add-annotations o [12:27:01] 
$ helm list --namespace default
NAME 	NAMESPACE   	REVISION	UPDATED                                	STATUS  	CHART        	APP VERSION
fleet	fleet-system	1       	2021-03-02 12:24:14.476796873 +0100 CET	deployed	fleet-0.3.300	0.3.3      
test 	default     	2       	2021-02-26 11:10:20.787363474 +0100 CET	deployed	empty-0.2.0  	           
test2	default     	1       	2021-02-26 11:08:09.799608082 +0100 CET	deployed	empty-0.1.0  	           

# itxaka @ zeus in ~/projects/hypper on git:viccuad/add-annotations o [12:27:08] 
$ helm list --namespace fleet-system
NAME	NAMESPACE	REVISION	UPDATED	STATUS	CHART	APP VERSION

```

Im not sure why this is...

**Comment by Itxaka** [comment]
kind of crazy, if we pass the `-n` flag to hypper then it identifies and sets the release in the proper helm namespace but...it deploys it in the default one in kubectl :thinking: 

```
# itxaka @ zeus in ~/projects/hypper on git:viccuad/add-annotations x [13:42:10] 
$ bin/hypper install --namespace fleet-system fleet ./fleet
Installing chart "fleet" in namespace "fleet-system"…
Done! 👏 

# itxaka @ zeus in ~/projects/hypper on git:viccuad/add-annotations x [13:42:27] 
$ kubectl get pods --all-namespaces                        
NAMESPACE            NAME                                         READY   STATUS    RESTARTS   AGE
default              fleet-controller-ccc95b8cd-4hp7k             0/1     Error     0          3s
default              gitjob-5997858b9c-bjdkw                      1/1     Running   0          3s
kube-system          coredns-f9fd979d6-7rb5b                      1/1     Running   0          8d
kube-system          coredns-f9fd979d6-r5d5p                      1/1     Running   0          8d
kube-system          etcd-kind-control-plane                      1/1     Running   0          8d
kube-system          kindnet-2bwg2                                1/1     Running   0          8d
kube-system          kube-apiserver-kind-control-plane            1/1     Running   0          8d
kube-system          kube-controller-manager-kind-control-plane   1/1     Running   1          8d
kube-system          kube-proxy-nq4gx                             1/1     Running   0          8d
kube-system          kube-scheduler-kind-control-plane            1/1     Running   0          8d
local-path-storage   local-path-provisioner-78776bfc44-x25nk      1/1     Running   1          8d          

# itxaka @ zeus in ~/projects/hypper on git:viccuad/add-annotations x [13:42:42] 
$ bin/hypper list --namespace fleet-system
NAME 	NAMESPACE   	REVISION	UPDATED                                	STATUS  	CHART        	APP VERSION
fleet	fleet-system	1       	2021-03-02 13:42:27.505382267 +0100 CET	deployed	fleet-0.3.300	0.3.3   
```

**Comment by Itxaka** [comment]
tested using the mongodb chart(https://github.com/bitnami/charts/tree/master/bitnami/mongodb) with an added annotation `catalog.cattle.io/namespace: mongospace` and it worked...on the kubectl side.

```
# itxaka @ zeus in ~/projects/hypper on git:viccuad/add-annotations x [13:52:56] 
$ kubectl get pods -n mongospace   
NAME                             READY   STATUS    RESTARTS   AGE
mongo-mongodb-6494f9b55d-j9xwg   1/1     Running   0          103s
```

on the helm/hypper side namespace shows correctly but its not really on the namespace mentioned, only on the default one:

```
# itxaka @ zeus in ~/projects/hypper on git:viccuad/add-annotations x [13:54:24] 
$ bin/hypper list                                             
NAME 	NAMESPACE   	REVISION	UPDATED                                	STATUS  	CHART         	APP VERSION
fleet	fleet-system	1       	2021-03-02 13:44:23.86441114 +0100 CET 	deployed	fleet-0.3.300 	0.3.3      
mongo	mongospace  	1       	2021-03-02 13:52:41.467714602 +0100 CET	deployed	mongodb-10.7.2	4.4.4      
test 	default     	2       	2021-02-26 11:10:20.787363474 +0100 CET	deployed	empty-0.2.0   	           
test2	default     	1       	2021-02-26 11:08:09.799608082 +0100 CET	deployed	empty-0.1.0   	           

# itxaka @ zeus in ~/projects/hypper on git:viccuad/add-annotations x [13:54:42] 
$ bin/hypper list -n mongospace           
NAME	NAMESPACE	REVISION	UPDATED	STATUS	CHART	APP VERSION
```

**Comment by viccuad** [comment]
@Itxaka you are right. @dragonchaser and I were debugging this, and it seems that Helm uses the `action.Install.Namespace` field for Helm's etcd database. And the chart's Kubernetes resources get installed depending on what is set in `action.Install.cfg.Kubeclient.Data.Namespace`.

In our case, Which in our case `action.cfg.Kubeclient.Data.Namespace` is `""` instead of the `fleet-system`. We seem to be setting `Install.Namespace` after `Install` has been used to set the `cfg.Kubeclient`.

**Comment by viccuad** [comment]
Thanks for the fix, Matt!

Tested it successfully for the 4 cases of namespace usage, with a vanilla and hypper chart alike.

**Review by Itxaka — APPROVED** [review]
very nice!

File: pkg/action/install.go [hunk 5]
```diff
@@ -1,43 +1,164 @@
...
+	// len(args) == 1
+	return args[0], nil
+}
+
+// NameAndChart overloads Helm's NameAndChart. It always fails.
+//
+// On Hypper, we need to read the chart annotations to know the correct release name.
+// Therefore, it cannot happen in this function.
+func (i *Install) NameAndChart(args []string) (string, string, error) {
+	return "", "", errors.New("NameAndChart() cannot be used")
```
dragonchaser, line 163: why not return

```golang
return nil, nil, errors.New("NameAndChart() cannot be used")
```

I do not see the point of having empty strings here since the error should be evaluated anyway in the code that uses this function. Also empty string might be valid when accidentely used and can cause follow-up issues.

cmd/hypper/install.go
```diff
@@ -78,6 +100,20 @@ func runInstall(args []string, client *action.Install, valueOpts *values.Options
 		return nil, err
 	}
 
+	if settings.NamespaceFromFlag {
+		client.Namespace = settings.Namespace()
+	} else {
+		client.SetNamespace(chartRequested, settings.Namespace())
+	}
```
dragonchaser, line 107: Could we have a helper package for this? I think this is needed in several places.

go.mod
```diff
@@ -11,8 +11,16 @@ require (
...
+	helm.sh/helm/v3 v3.5.2
+	k8s.io/cli-runtime v0.20.4
+)
+
+replace (
+	// WARNING! Do NOT replace these without also replacing their lines in the `require` stanza below.
+	// These `replace` stanzas are IGNORED when this is imported as a library
+	github.com/docker/distribution => github.com/docker/distribution v0.0.0-20191216044856-a8371794149d
+	github.com/docker/docker => github.com/moby/moby v17.12.0-ce-rc1.0.20200618181300-9dc6525e6118+incompatible
+	github.com/mattfarina/hypper => github.com/rancher-sandbox/hypper v0.0.0-20210218091504-4d1470834315
```
dragonchaser, line 25: this line needs to go

pkg/cli/environment.go
```diff
@@ -131,6 +149,8 @@ func envCSV(name string) (ls []string) {
 	return
 }
 
+// EnvVars reads the hypper and helm enviromental variables and populates
+// values in the EnvSettings caller.
 func (s *EnvSettings) EnvVars() map[string]string {
 	envvars := map[string]string{
 		"HELM_BIN":               os.Args[0],
```
dragonchaser, line 156: What is this used for?

dragonchaser (changes requested):

cmd/hypper/install.go
```diff
@@ -18,18 +20,32 @@ import (
...
 	valueOpts := &values.Options{}
+	var outfmt output.Format
+
 	cmd := &cobra.Command{
 		Use:   "install [NAME] [CHART]",
 		Short: "install a chart",
 		Long:  installDesc,
 		Args:  require.MinimumNArgs(1),
 		RunE: func(_ *cobra.Command, args []string) error {
-			logger.Info(eyecandy.ESPrintf(settings.NoEmojis, ":cruise_ship: Installing %s…", args[0]))
```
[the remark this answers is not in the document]
viccuad: It was substituted by this line, so we could list `ReleaseName` and `Namespace`: https://github.com/rancher-sandbox/hypper/pull/49/files#diff-5d4849bad97d2e075ce5d0ae5af3e93dbfde235ce1345fa96148279cf706af2fR61
I dropped printing an emoticon on that exact line to not depend on pkg/eyecandy inside the library calls yet. I'm still wondering if the logger should be the one calling the eyecandy functions itself.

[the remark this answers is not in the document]
dragonchaser: But no emoticons there :disappointed:, just kidding, but I think this was one of the requirements to have emoticons there, I personally can live without that.

pkg/action/install.go
viccuad: well, the default value for a string is `""`, not `nil`.  The idea of this function overload is to fail always to deactivate this function. This is one function that we obtained from Helm's type composition, but it can't return correct values, as in Hypper we need to look at the chart annotations, and the signature of this function doesn't get a chart.

dragonchaser: Yes, but if you return an error anyway, returning nil for the expected values is often used. So I'd prefer that.

cmd/hypper/install.go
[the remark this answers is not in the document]
viccuad: I bent the requirements a bit :). They said: "emoticons when `hypper install`" and there are. Just not at that specific time.

pkg/action/install.go
viccuad: There are no nil strings in go, so I don't think one can do what you suggest. See for Helm for example:
https://github.com/helm/helm/blob/b30f61d76de3805a839a95dc0f8ef8d8e208b09f/pkg/action/install.go#L564
https://github.com/helm/helm/blob/b30f61d76de3805a839a95dc0f8ef8d8e208b09f/pkg/action/install.go#L638
and others.

pkg/cli/environment.go
viccuad: So far, it's only used for resetting the environment after tests, in https://github.com/rancher-sandbox/hypper/blob/viccuad%2Fadd-annotations/pkg/cli/environment_test.go#L120.
It seems that it would be used more in plugins and cmd/helm/env.go.

cmd/hypper/install.go
viccuad: :thinking:  I thought about that, providing this logic in a library function. Thing is, that the Helm library methods and functions expect the user to find out which `ReleaseName` and `Namespace` to use on their own, and once they know them, call the library functions with them. So I followed the same pattern.

This is because it is Hypper's cli the one that has the priorities of "first --namespace flag, then namespace annotations, then default namespace in kube context". Someone else using the Hypper libraries could decide on different priorities. Hence why in my opinion this shouldn't be provided on hypper libs.

We could provide it on a cmd/hypper function, but since I have yet to find a second usage for it, I felt it as is.

dragonchaser: Sounds reasonable.

go.mod
viccuad: Good catch! fixed in https://github.com/rancher-sandbox/hypper/pull/49/commits/84ccd464906e0a156518bf7b45f91cc1b9c3f491.

pkg/cli/environment.go
dragonchaser: :+1:

dragonchaser (approved):

Itxaka:
rebase required!

Itxaka (approved):

PR #51, opened 2021-02-24
add repo commands (old)

dragonchaser:


.golangci.yml
```diff
@@ -0,0 +1,26 @@
+run:
+  timeout: 2m
+
+linters:
+  disable-all: true
```
viccuad, line 5: I don't like disabling `errcheck` (and others) in golangci-lint.
I would very much prefer for _internal/registry_ to contain unchanged helm code (if we need changes, we overload by composition in _hypper/pkg_).

Once we have _internal/registry_ with unchanged helm code, we can just ignore that folder with `run.skip-dirs` (see https://golangci-lint.run/usage/configuration/#config-file).

viccuad (changes requested):

viccuad:
Commit https://github.com/rancher-sandbox/hypper/pull/51/commits/f1fd045aca0522d1c7211a855bb35cad3e47226d adds 14000 lines of code, and ~150 lines. I think this is wrong. How does one know which code is ours and which is Helms? From what point in Helm's history is the code (current master, v3.3, ...)? If any code gets copy-pasted from Helm, it should at minimum be contained to 1 commit, which message specifies the Helm' history point from which it was copy-pasted (I would prefer git subtrees, etc).

Also, commit https://github.com/rancher-sandbox/hypper/pull/51/commits/f1fd045aca0522d1c7211a855bb35cad3e47226d drops a lot of Helm packages in _internal/helm/pkg/_, for example _internal/helm/pkg/{getter,repo}_. The same concerns arise.
I would very much give a try using to composition instead, and any hypper modifications such be simple and in _hypper/pkg/{getter,repo}_.

There's also  _internal/helm/plugin_ (and maybe others) that should go in _internal/helm/_ by your approach. But I think this is not sustainable, and it's worth a try to just use composition instead.

pkg/hypperpath/lazypath.go
```diff
@@ -0,0 +1,59 @@
...
+	// for the cache directory. When no value is set a default is used.
+	CacheHomeEnvVar = "HYPPER_CACHE_HOME"
+
+	// ConfigHomeEnvVar is the environment variable used by Helm
+	// for the config directory. When no value is set a default is used.
+	ConfigHomeEnvVar = "HYPPER_CONFIG_HOME"
+
+	// DataHomeEnvVar is the environment variable used by Helm
+	// for the data directory. When no value is set a default is used.
+	DataHomeEnvVar = "HYPPER_DATA_HOME"
```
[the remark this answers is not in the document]
dragonchaser: Probably, but I do not like the idea, there will be cases where the hypper settings are being overturned by helm stuff consumed through comppositing and this will probably screw up the local helm config of the person running this.

viccuad:
Also, we need to provide a _hypper/pkg/repo_ analogous to _helm/pkg/repo_, as we are offering hypper as a library.

.golangci.yml
mattfarina: @viccuad I like your idea of using skip-dirs on imported 3rd party code.

Issue #52, opened 2021-02-25
hypper status command


mattfarina:
A `hypper status` command that tells you the status of a running instance of a chart.

PR #53, opened 2021-02-26
Add status command

Itxaka:
Basically lifted from the helm status command with minor changes:
 - Uses log-go for output
 - Removes tests for auto completion as we dont have that yet

Fixes: #52 

### DIFF
--- a/cmd/hypper/status.go
+++ b/cmd/hypper/status.go
@@ -1,0 +1,178 @@
+/*
+Copyright The Helm Authors, SUSE
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"fmt"
+	"github.com/Masterminds/log-go"
+	logio "github.com/Masterminds/log-go/io"
+	"github.com/rancher-sandbox/hypper/pkg/action"
+	"github.com/spf13/cobra"
+	"helm.sh/helm/v3/cmd/helm/require"
+	"helm.sh/helm/v3/pkg/chartutil"
+	"helm.sh/helm/v3/pkg/cli/output"
+	"helm.sh/helm/v3/pkg/release"
+	"io"
+	"strings"
+	"time"
+)
+
+// NOTE: Keep the list of statuses up-to-date with pkg/release/status.go.
+var statusHelp = `
+This command shows the status of a named release.
+The status consists of:
+- last deployment time
+- k8s namespace in which the release lives
+- state of the release (can be: unknown, deployed, uninstalled, superseded, failed, uninstalling, pending-install, pending-upgrade or pending-rollback)
+- revision of the release
+- description of the release (can be completion message or error message, need to enable --show-desc)
+- list of resources that this release consists of, sorted by kind
+- details on last test suite run, if applicable
+- additional notes provided by the chart
+`
+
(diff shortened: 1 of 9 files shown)


cmd/hypper/status.go
```diff
@@ -0,0 +1,177 @@
...
+
+			// strip chart metadata from the output
+			rel.Chart = nil
+
+			return outfmt.Write(wInfo, &statusPrinter{rel, false, client.ShowDescription})
+		},
+	}
+
+	f := cmd.Flags()
+
```
viccuad, line 74: nitpick: we could add the `--revision` and `--description` flags here, as Helm does

pkg/action/status.go
```diff
@@ -0,0 +1,48 @@
...
+package action
+
+import (
+	"helm.sh/helm/v3/pkg/release"
+)
+
+// Status is the action for checking the deployment status of releases.
+//
+// It provides the implementation of 'helm status'.
+type Status struct {
```
viccuad, line 23: The Status type could be a composite type of helm/pkg/status.Status, with cfg overloaded to our own Configuration type. That way new fields of helm/pkg/status.Status besides Version and ShowDescription would appear for us automatically. And we could reuse helm/pkg/status.Run().

But maybe for this simple type is more hassle than anything.

viccuad (dismissed):

cmd/hypper/status.go
Itxaka: great point, I missed that flag!

pkg/action/status.go
Itxaka: Nice idea, I have no idea how to do that but seems a good opportunity to learn!

cmd/hypper/status.go
Itxaka: should be fixed now!

pkg/action/status.go
Itxaka: done, thanks for the hint!

pkg/action/status.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"helm.sh/helm/v3/pkg/action"
+)
+
+// Status is the action for checking the deployment status of releases.
+//
+// It provides the implementation of 'helm status'.
+type Status struct {
+	*action.Status
```
viccuad, line 24: I think that as it is, we are using Helm's `Configuration` type inside of the embedded `helm/pkg/action.Status` struct. But we have our own Hypper's `Configuration` type that we should be using. One needs to overload the `cfg` here:

```suggestion
	*action.Status
	 cfg *Configuration
```

If we don't overload, it may fail somewhere. We don't have unit tests on `NewStatus()` (Helm doesn't, either, because it seems to be a simple enough pkg). But at some point it would have failed.

viccuad (changes requested):
Nice and fast PR that looks good :).

pkg/action/status.go
mattfarina: As much as we can we should directly use the Helm packages as packages rather than copying code over. This way we inherit updates such as bug and security fixes.

viccuad: For what is worth, our `Configuration` type is also a composite type of Helm's `Configuration` type, exactly for that reason: to import Helm code and not copy it.

Itxaka: Dont we do the same on the `List` part? What is the difference?

Also, when calling `NewStatus` arent we passing our `Configuration` found at `hypper/pkg/action/action.go` which is a composite of our+helm one? I agree that using first helm config and then overloading ours is not ideal, as that could lead to bugs.

Like, I can change this, but we should agree on how to deal with this to make it standard across our code.

viccuad: @Itxaka  I think you are right, and actually List should be implemented as:

```
type List struct {
	*action.List
	cfg *Configuration
}
```

Aaaand that was an oversight from me when I implemented it. We should fix that in a follow-up PR :).

dragonchaser: Also we should not access the underlying helm code directly in a composition, this can and will cause weird problems, if needed overload and access indirectly through our own types.

viccuad (approved):

dragonchaser (approved):

Issue #54, opened 2021-03-01
Hypper upgrade command


mattfarina:
This is similar to the `helm upgrade` command with the different being that it does not need the release name or namespace passed to it if specified in the chart. Depends on #49.

PR #58, opened 2021-03-02
Add repo commands

dragonchaser:
This PR adds the following commands to `hypper`:
- `hypper repo add`
- `hypper repo index`
- `hypper repo list`
- `hypper repo update`
- `hypper repo delete`

It imports `internal/helm/urlutil` from Helm to `internal/third-party/helm/urlutil`. `internal/third-party` gets automatically skipped by golangci-lint et al.
It refactors the `pkg/cli/environment.go.EnvSettings` type into a composite type of Helm's, and then overloads it's fields to make sure nothing uses Helm's setting values. TODO use reflection to check that all fields are overloaded, see https://github.com/rancher-sandbox/hypper/issues/65.
It adds `internal/test/ensure/*`.
It adds `pkg/{hypperpath,repo}`.

Supersedes #51
Closes https://github.com/rancher-sandbox/hypper/issues/26

PR #64, opened 2021-03-03
Add new upgrade command

Itxaka:
Lifted completely from helm, no changes other that removing
autocompletion as we dont have that yet

fixes: #54


cmd/hypper/upgrade.go
```diff
@@ -0,0 +1,193 @@
+/*
+Copyright The Helm Authors.
```
dragonchaser, line 2: It looks like there are changes from the helm code here, can you please add SUSE to the copyright.
e.g. 
```
Copyright The Helm Authors, SUSE.
```

That way it is easier for us to track changed code in maintenance.

cmd/hypper/upgrade.go
```diff
@@ -0,0 +1,193 @@
...
+	var createNamespace bool
+
+	cmd := &cobra.Command{
+		Use:   "upgrade [RELEASE] [CHART]",
+		Short: "upgrade a release",
+		Long:  upgradeDesc,
+		Args:  require.ExactArgs(2),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			client.Namespace = settings.Namespace()
+			helmSettings := helmCli.New()
```
dragonchaser, line 79: Please do not create an instance of the helmCli manually, this includes code that accesses helms environment variables and can cause severe problems if a user runs both `hypper` and `helm` on the same machine.
We have introduced this as composition in https://github.com/rancher-sandbox/hypper/blob/handle_repos_composition/pkg/cli/environment.go#L40 where we try to make sure to overwrite those, please use this as soon as PR #58 is merged.

internal/test/test.go
```diff
@@ -104,3 +105,13 @@ func update(filename string, in []byte) error {
 func normalize(in []byte) []byte {
 	return bytes.Replace(in, []byte("\r\n"), []byte("\n"), -1)
 }
+
+// TempDir ensures a scratch test directory for unit testing purposes.
+func TempDir(t *testing.T) string {
+	t.Helper()
+	d, err := ioutil.TempDir("", "helm")
```
dragonchaser, line 112: `s/helm/hypper/`

dragonchaser (changes requested):

cmd/hypper/upgrade.go
Itxaka: I was currently looking at that part of your PR, will do as soon as its merged!

cmd/hypper/upgrade.go
Itxaka: I dont think there are any changes to helm code here...yet.

On my next commit there should be changes and I will add that.

cmd/hypper/upgrade.go
dragonchaser: Thank you :)

cmd/hypper/upgrade.go
dragonchaser: :+1:

kkaempf: That should be "SUSE LLC" :wink:

internal/test/test.go
Itxaka: This is actually removed as your PR brings the same function, so I will drop this.

cmd/hypper/dependency_build.go
```diff
@@ -0,0 +1,15 @@
+package main
```
viccuad, line 1: cmd/hypper/dependency_build.go is missing a header.

cmd/hypper/upgrade_test.go
```diff
@@ -0,0 +1,423 @@
...
+
+	relWithStatusMock := func(n string, v int, ch *chart.Chart, status release.Status) *release.Release {
+		return release.Mock(&release.MockReleaseOptions{Name: n, Version: v, Chart: ch, Status: status})
+	}
+
+	relMock := func(n string, v int, ch *chart.Chart) *release.Release {
+		return release.Mock(&release.MockReleaseOptions{Name: n, Version: v, Chart: ch})
+	}
+
+	tests := []cmdTestCase{
```
viccuad, line 103: I suppose in the future (not on this PR) we need to start adding more tests for hypper charts in all commands and subcomands, testing the upgrade etc when there's shared deps, and so on :grimacing:.

cmd/hypper/upgrade.go
```diff
@@ -0,0 +1,196 @@
...
+	"helm.sh/helm/v3/pkg/cli/output"
+	"helm.sh/helm/v3/pkg/cli/values"
+	"helm.sh/helm/v3/pkg/getter"
+	"helm.sh/helm/v3/pkg/storage/driver"
+)
+
+const upgradeDesc = `
+This command upgrades a release to a new version of a chart.
+
+The upgrade arguments must be a release and chart. The chart
```
viccuad, line 40: "The upgrade arguments must be a release and chart": cmd lists `upgrade [CHART]`

cmd/hypper/upgrade.go
```diff
@@ -0,0 +1,196 @@
...
+or use the '--set' flag and pass configuration from the command line, to force string
+values, use '--set-string'. In case a value is large and therefore
+you want not to use neither '--values' nor '--set', use '--set-file' to read the
+single large value from file.
+
+You can specify the '--values'/'-f' flag multiple times. The priority will be given to the
+last (right-most) file specified. For example, if both myvalues.yaml and override.yaml
+contained a key called 'Test', the value set in override.yaml would take precedence:
+
+    $ helm upgrade -f myvalues.yaml -f override.yaml redis ./redis
```
viccuad, line 55: Same here, the docs don't match the command:

```suggestion
    $ helm upgrade -f myvalues.yaml -f override.yaml ./redis
```

Also, I'm not sure how all works, given that we are setting `Args:  require.ExactArgs(1),`, which means we only accept chart and not release-name.

cmd/hypper/upgrade.go
```diff
@@ -0,0 +1,196 @@
...
+			if ch.Metadata.Deprecated {
+				logger.Warn(eyecandy.ESPrint(settings.NoEmojis, "This chart has been deprecated :exclamation:"))
+
+			}
+			if client.ReleaseName == "" {
+				client.ReleaseName, _ = client.Name(ch, args)
+				logger.Debugf("client.ReleaseName was empty, setting release name to %s", client.ReleaseName)
+			}
+
+			// Fixes #7002 - Support reading values from STDIN for `upgrade` command
```
viccuad, line 105: nitpick (maybe not needed, as it complicates upstreaming?)

```suggestion
			// Fixes helm#7002 - Support reading values from STDIN for `upgrade` command
```

cmd/hypper/upgrade.go
```diff
@@ -0,0 +1,196 @@
...
+			// Must load values AFTER determining if we have to call install so that values loaded from stdin are are not read twice
```
viccuad, line 106: If I understand correctly comment specifies that the `if client.Install {}` *needs* to run before loading values. If not, values are read twice.

Which means that this block should come first thing in the `RunE: Func()` as it is in Helm.

cmd/hypper/upgrade.go
```diff
@@ -0,0 +1,196 @@
...
+
+func newUpgradeCmd(cfg *action.Configuration, logger log.Logger) *cobra.Command {
+	client := action.NewUpgrade(cfg)
+	valueOpts := &values.Options{}
+	var outfmt output.Format
+	var createNamespace bool
+
+	cmd := &cobra.Command{
+		Use:   "upgrade [CHART]",
+		Short: "upgrade a release",
```
viccuad, line 72:
```suggestion
		Short: "upgrade a chart",
```

viccuad (changes requested):

pkg/action/history.go
```diff
@@ -0,0 +1,58 @@
...
+	"helm.sh/helm/v3/pkg/release"
+)
+
+// History is the action for checking the release's ledger.
+//
+// It provides the implementation of 'helm history'.
+// It returns all the revisions for a specific release.
+// To list up to one revision of every release in one specific, or in all,
+// namespaces, see the List action.
+type History struct {
```
viccuad, line 32: We are reimplementing `helm/pkg/action.History` by copy-pasting it into `hypper/pkg/action.History`. I would prefer it used a composite type here, and its constructor instead, as it gives us forward compatibility.

viccuad (changes requested):

pkg/action/upgrade.go
```diff
@@ -0,0 +1,43 @@
...
+
+// NewUpgrade creates a new Upgrade object with the given configuration.
+func NewUpgrade(cfg *Configuration) *Upgrade {
+	return &Upgrade{
+		Upgrade: action.NewUpgrade(cfg.Configuration),
+		cfg:     cfg,
+	}
+}
+
+func (i *Upgrade) Name(chart *chart.Chart, args []string) (string, error) {
```
viccuad, line 24: Missing doc comment on Name(), as it is exported (shouldn't `golangci-lint` catch this? It hasn't on CI Oo).

Silly question, can we just use `pkg/action/install.go.Name()` somehow? https://github.com/rancher-sandbox/hypper/blob/main/pkg/action/install.go#L83-L140

If not, this should be a static function, not a method, as I don't see it making use of the action in  `i`. If it were a static function, I would put it in a separate file in pkg/action/. But it would be nice if we didn't have this logic duplicated.

viccuad (changes requested):

cmd/hypper/upgrade_test.go
Itxaka: :+1: 

Agreed, once this settles we need to start adding all types of tests to cover for more.

cmd/hypper/dependency_build.go
Itxaka: :+1:

cmd/hypper/upgrade.go
Itxaka: :+1:

cmd/hypper/upgrade.go
Itxaka: :+1:

cmd/hypper/upgrade.go
Itxaka: Good point, I like it!

cmd/hypper/upgrade.go
Itxaka: This actually refers to line 149 in the same file, I went through the mentioned issue and PR and the problem was that before they were loading the `vals` first and then calling `client.Install` which would load the `vals` again, thus reading the `vals` twice.

`vals, err := valueOpts.MergeValues(getter.All(settings.EnvSettings))`

Now that the `vals` are below the `client.Install` they are only read before calling `client.Upgrade`, and we leave `client.Install` to load the `vals` itself.

cmd/hypper/upgrade.go
Itxaka: :+1:

pkg/action/upgrade.go
Itxaka: > Silly question, can we just use pkg/action/install.go.Name() somehow? https://github.com/rancher-sandbox/hypper/blob/main/pkg/action/install.go#L83-L140

This would need us instantiating the Install with all the params which is cumbersome and could hide some issues.

> If not, this should be a static function, not a method, as I don't see it making use of the action in i. If it were a static function, I would put it in a separate file in pkg/action/. But it would be nice if we didn't have this logic duplicated.

So I thought about this and Im conflicted.

On one hand, I think we should move this into a `chartutils` package as its common to both
On the other, we may need to do different things to obtain the name in the `Install` vs the `Upgrade`

I lean more heavily into moving it to chartutils, but would like more feedback.

pkg/action/history.go
Itxaka: woops, good point. I just bindly copied that there trying to avoid more complexity, let me change that!

cmd/hypper/upgrade.go
viccuad: Nice! many thanks for the explanation and sorry for the noise! :)

pkg/action/upgrade.go
viccuad: :thinking: mmh, true, good point!.
So, both Install and Upgrade are of type Action, so if we need access to Namespace and ReleaseName, both will have it in case we need it in the future (we already need them in `Install.Name()`).

How about creating a new unexported function in pkg/action/name.go (or other), containing the logic on https://github.com/rancher-sandbox/hypper/blob/main/pkg/action/install.go#L108-L115, that both `pkg/action/install.Name()` and `pkg/action/Upgrade.Name()` call?
I would prefer if that new unexported function contained the following logic: https://github.com/rancher-sandbox/hypper/blob/main/pkg/action/install.go#L100-L115 but the amount of flags is different between Install and Upgrade.

internal/test/test.go
```diff
@@ -19,10 +19,9 @@ package test
 import (
 	"bytes"
 	"flag"
+	"github.com/pkg/errors"
```
dragonchaser, line 22: this is a formatting thing of go vet, should this be commited?

Itxaka: formatting indeed,this is `make test` autoformatting it via `go vet`

I think we commit it, otherwise its gonna end up there sooner or later because we will always run make test before committing....

pkg/action/upgrade.go
Itxaka: Do you mind if we do this change afterwards? That way I can get this out of the way and experiment with it later.

internal/test/test.go
dragonchaser: Ok

dragonchaser (approved):

pkg/action/upgrade.go
viccuad: Ok, let's do that. But then, let's open an issue to deduplicate the Name() code, or I for sure will forget :).

viccuad (approved):
Approving, let's open an issue for Name() deduplication.

Itxaka:
https://github.com/rancher-sandbox/hypper/issues/70

PR #66, opened 2021-03-04
copyright/license check

### LINKED ISSUE #8: Add missing License and Copyright headers

dragonchaser:
This is a rudimentary checker for `make` or ci that checks if go files start with the `package` keyword, if it warns about (probable) missing license/copyright headers.

- Adds `make license-check` target

Closes #8

### DIFF
--- a/.github/workflows/license-check.yml
+++ b/.github/workflows/license-check.yml
@@ -1,0 +1,20 @@
+name: license-check
+on:
+  push:
+    tags:
+      - v*
+    branches:
+      - main
+  pull_request:
+jobs:
+  license-check:
+    strategy:
+      matrix:
+        go-version: ["1.16.x"]
+        os: [ubuntu-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Check license
+        run: make license-check
--- a/.github/workflows/license-check.yml
+++ b/.github/workflows/license-check.yml
@@ -8,11 +8,7 @@ on:
   pull_request:
 jobs:
   license-check:
-    strategy:
-      matrix:
-        go-version: ["1.16.x"]
-        os: [ubuntu-latest]
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-latest
     steps:
       - name: Checkout
         uses: actions/checkout@v2
--- a/scripts/license_check.sh
+++ b/scripts/license_check.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+
+FILES=$(find . -type f -name "*.go"|grep -v 'third_party'|grep -v 'testdata')
+FAIL=0
+
+for file in  $FILES; do
+    error=""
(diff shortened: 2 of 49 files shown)


scripts/license_check.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+FILES=$(find . -type f -name "*.go")
```
viccuad, line 3: We should skip `testdata`, `third-party` dirs ala [helm](https://github.com/helm/helm/blob/master/scripts/validate-license.sh).

scripts/license_check.sh
```diff
@@ -0,0 +1,16 @@
...
+FAIL=0
+for file in  $FILES; do
+    if [ $(head -n1 "${file}"|grep package|wc -l) -eq 0 ]; then
```
viccuad, line 6: I would like to check for the specific license  that we use, as Helm does (also, the script is called license_check.sh but doesn't check licenses).

I would also prefer to check for either Rancher, SUSE or Rancher & SUSE copyright. As it is, leaving a newline on the beginning of the file will pass the lint.

viccuad:
Uoook, just realized that the PR is WIP :sweat_smile: . Please feel free to dismiss my review, sorry!

scripts/license_check.sh
dragonchaser: Sounds good...

scripts/license_check.sh
dragonchaser: Makes sense, I will see if this can be invoked through github itself, this file is primarily intended to check for missing/wrong headers.

.github/workflows/license-check.yml
```diff
@@ -0,0 +1,20 @@
...
+  push:
+    tags:
+      - v*
+    branches:
+      - main
+  pull_request:
+jobs:
+  license-check:
+    strategy:
+      matrix:
```
Itxaka, line 12: This should not be needed as this is a simple bash script and does not require a matrix of versions or os to run on

Itxaka (changes requested):

.github/workflows/license-check.yml
dragonchaser: Makes sense, will change.

scripts/license_check.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+
+FILES=$(find . -type f -name "*.go"|grep -v 'third_party'|grep -v 'testdata')
```
viccuad, line 3: We may want to expand `FILES` to contain more than .go files. Maybe just listing the files of the prpject, besides third_party, testdata or others. Because `scripts/license_check.sh` is missing a license header, and it's not being detected.

dragonchaser: We might want, I think using only .go atm is sufficient. Also I do not see a point in adding license data in this trivial script. It does not contain anything worthy like business logic etc.

Itxaka (approved):

PR #69, opened 2021-03-08
Create temp dir with hypper pattern

Itxaka:
Instead of using helm for the pattern


viccuad (approved):
Wooops! Nice, good catch!

Issue #68, opened 2021-03-07
helm install option `--values` not recognized


kkaempf:
I'm trying to use hypper as a drop-in replacement for helm.

This helm command works:

> &gt; helm install chimera-controller chimera/chimera-controller --values custom-values.yaml

Replacing `helm` with `hypper` it doesn't work:

> &gt; hypper install chimera-controller chimera/chimera-controller --values custom-values.yaml
Error: unknown flag: --values

Itxaka:
We are missing some flags, I noticed this while calling `install` from the upgrade, let me do a quick PR

kkaempf:
fix confirmed. Thanks !

PR #71, opened 2021-03-08
Add missing flag to Install

Itxaka:
When mimicking the addInstallFlags from Install we
missed the values flags and the chartoptions flags

Fixes: #68 

### DIFF
--- a/cmd/hypper/install.go
+++ b/cmd/hypper/install.go
@@ -70,7 +70,10 @@ func newInstallCmd(actionConfig *action.Configuration, logger log.Logger) *cobra
 			return nil
 		},
 	}
-	addInstallFlags(cmd, cmd.Flags(), client, valueOpts)
+	f := cmd.Flags()
+	addInstallFlags(cmd, f, client, valueOpts)
+	addValueOptionsFlags(f, valueOpts)
+	addChartPathOptionsFlags(f, &client.ChartPathOptions)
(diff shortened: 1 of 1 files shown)


Itxaka:
license check failing due hto ttps://github.com/rancher-sandbox/hypper/pull/72

kkaempf:
Is there a better way to implement this ? We would need to update hypper every time helm adds a new flag. Can't we just pass all (unknown) flags to helm ?

Itxaka:
> Is there a better way to implement this ? We would need to update hypper every time helm adds a new flag. Can't we just pass all (unknown) flags to helm ?

On `Upgrade` we are using the helm flags directly without modifying them in our side. In this case I think its because we need to add a flag that collides with the flag passed from Helm (Due to hypper skipping the release name as a feature)

I think we should open another ticket to deal with the flag stuff as we may be able to bring Helm flags and then force-override ours on top somehow.

dragonchaser:
> Is there a better way to implement this ? We would need to update hypper every time helm adds a new flag. Can't we just pass all (unknown) flags to helm ?

I don't think this is feasible, we also might include changes to our own values in the annotated charts, so we have to evaluate them anyhow.

viccuad (approved):

viccuad:
Thanks for looking into this :)! Good to have usage parity with Helm.

dragonchaser (approved):

PR #72, opened 2021-03-08
Add missing action/upgrade.go license

Itxaka:
license checker was failing due merging 2 prs at the same
time, one that dixed all license headers and the other
that introduced this file, so the license checker was
green until now :D

### DIFF
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -1,3 +1,18 @@
+/*
+Copyright The Helm Authors, SUSE LLC.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


dragonchaser (approved):

pkg/action/upgrade.go
```diff
@@ -1,3 +1,18 @@
+/*
+Copyright The Helm Authors, SUSE LLC.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
```
mattfarina, line 8: The URL should not be indented. We have it all the way on the left in the other files.

mattfarina (changes requested):
Good catch on the missing header. One minor nit...

pkg/action/upgrade.go
Itxaka: we dont because I copied it from https://github.com/rancher-sandbox/hypper/blob/main/pkg/action/install.go#L8 lol

viccuad: It seems that some one them are indented https://github.com/rancher-sandbox/hypper/blob/main/pkg/action/install_test.go#L8
and some of them are not https://github.com/rancher-sandbox/hypper/blob/main/pkg/action/action.go#L7

viccuad (approved):

Issue #73, opened 2021-03-08
hypper upgrade not using namespace from annotations


mattfarina:
My test setup is with the fleet chart. Here's an example of how to get it...

```sh
$ wget https://github.com/rancher/charts/raw/dev-v2.5/released/assets/fleet/fleet-0.3.000.tgz
```

Then I use hypper to install the chart...

```sh
$ kubectl create ns fleet-system
$ hypper install fleet-0.3.000.tgz
```

After this you can see the release name is fleet and the resource were installed into the fleet-system namespace.

Then do an upgrade...

```sh
$ hypper upgrade fleet-0.3.000.tgz
```

This caused the revision to increment from 1 to 2 for the release. But, the upgraded resources are now in the current namespace and the original resources are still in the fleet-system namespace. If you uninstall fleet one set of resources is left behind, too.

The namespace from the annotations should be used on upgrade, too.

Itxaka:
:+1 :+1: Can confirm. I had a comit for this but I probbly trashed by rebasing...

Issue #70, opened 2021-03-08
Fix name and namespace setting deduplication


Itxaka:
In both the install and upgrade commands we need to add a function `Name()` to calculate the name based of the annotations.

Lets try to minimize duplicates and consolidate those generic functions that are more general into one place. Also we avoid fixing bugs in 2 places!

viccuad:
The same will be true for dealing with Namespace logic and annotations. There's https://github.com/rancher-sandbox/hypper/blob/main/pkg/action/install.go#L71-L80, and I will need to repeat that in another action, SharedDependency.

PR #74, opened 2021-03-09
Fix namespace set on upgrade command

Itxaka:
Namespace was not being set via annotations properly
for the upgrade command nor the install command called
from the upgrade

Fixes: #73 

### DIFF
--- a/cmd/hypper/hypper_test.go
+++ b/cmd/hypper/hypper_test.go
@@ -43,7 +43,7 @@ import (
 func testTimestamper() time.Time { return time.Unix(242085845, 0).UTC() }
 
 func init() {
-	action.Timestamper = testTimestamper
+	helmAction.Timestamper = testTimestamper
 }
 
 func runTestCmd(t *testing.T, tests []cmdTestCase) {
--- a/cmd/hypper/upgrade.go
+++ b/cmd/hypper/upgrade.go
@@ -42,6 +42,14 @@ argument can be either: a chart reference('example/mariadb'), a path to a chart 
 a packaged chart, or a fully qualified URL. For chart references, the latest
 version will be specified unless the '--version' flag is set.
 
+There are four different ways you can select the release name and namespace
+where the chart will be upgraded. By priority order:
+
+1. By the args passed from the CLI: hypper upgrade --release-name mariadb example/mariadb --namespace system
+2. By using hypper.cattle.io annotations in the Chart.yaml
+3. By using catalog.cattle.io annotations in the Chart.yaml
+4. By using the current namespace as configured with the kubeconfig
+
 To override values in a chart, use either the '--values' flag and pass in a file
 or use the '--set' flag and pass configuration from the command line, to force string
 values, use '--set-string'. In case a value is large and therefore
@@ -98,7 +106,7 @@ func newUpgradeCmd(cfg *action.Configuration, logger log.Logger) *cobra.Command 
 
 			}
 			if client.ReleaseName == "" {
-				client.ReleaseName, _ = client.Name(ch, args)
+				client.ReleaseName, _ = client.Name(ch)
 				logger.Debugf("client.ReleaseName was empty, setting release name to %s", client.ReleaseName)
 			}
 
@@ -114,6 +122,15 @@ func newUpgradeCmd(cfg *action.Configuration, logger log.Logger) *cobra.Command 
 						logger.Info(eyecandy.ESPrintf(settings.NoEmojis, "Release %q does not exist. Installing it now. :toolbox:\n", client.ReleaseName))
 					}
 					instClient := action.NewInstall(cfg)
+					// Set namespace for the install client
+					if settings.NamespaceFromFlag {
+						instClient.Namespace = settings.Namespace()
+					} else {
+						instClient.SetNamespace(ch, settings.Namespace())
+					}
+					// By this time client.Namespace has the correct namespace
+					instClient.Config.SetNamespace(instClient.Namespace)
+
 					instClient.CreateNamespace = createNamespace
 					instClient.ChartPathOptions = client.ChartPathOptions
 					instClient.DryRun = client.DryRun
@@ -150,6 +167,16 @@ func newUpgradeCmd(cfg *action.Configuration, logger log.Logger) *cobra.Command 
 			if err != nil {
 				return err
 			}
+
+			// Set namespace for the upgrade client
+			if settings.NamespaceFromFlag {
+				client.Namespace = settings.Namespace()
+			} else {
+				client.SetNamespace(ch, settings.Namespace())
+			}
+
+			// By this time client.Namespace has the correct namespace, so set it to the storage
+			client.Config.SetNamespace(client.Namespace)
 
 			rel, err := client.Run(client.ReleaseName, ch, vals)
 			if err != nil {
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -17,16 +17,14 @@ limitations under the License.
 package action
 
 import (
-	"github.com/pkg/errors"
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/chart"
-	"strings"
 )
 
 // Upgrade is a composite type of Helm's Upgrade type
 type Upgrade struct {
 	*action.Upgrade
-	cfg         *Configuration
+	Config      *Configuration
 	ReleaseName string
 }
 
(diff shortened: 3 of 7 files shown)


cmd/hypper/upgrade.go
```diff
@@ -114,6 +122,12 @@ func newUpgradeCmd(cfg *action.Configuration, logger log.Logger) *cobra.Command
 						logger.Info(eyecandy.ESPrintf(settings.NoEmojis, "Release %q does not exist. Installing it now. :toolbox:\n", client.ReleaseName))
 					}
 					instClient := action.NewInstall(cfg)
+					// Set namespace for the install client
```
dragonchaser, line 125: we have almost identical code in three places now (twice here line 125-130, 168-174 and in install.go) can we move that to a static helper function?

cmd/hypper/upgrade.go
```diff
@@ -151,6 +165,13 @@ func newUpgradeCmd(cfg *action.Configuration, logger log.Logger) *cobra.Command
 				return err
 			}
 
+			// Set namespace for the upgrade client
```
dragonchaser, line 168: see previous comment

dragonchaser (changes requested):

cmd/hypper/upgrade.go
viccuad: I agree. I would schedule that inside https://github.com/rancher-sandbox/hypper/issues/70 (and rename the card to "fix name and namespace setting deduplication" or so).

viccuad: I would prefer to see all duplications everywhere, to see the pattern, and then fix them in a way that is good for all callers, hence why I want to postpone that.

cmd/hypper/upgrade.go
```diff
@@ -42,6 +42,14 @@ argument can be either: a chart reference('example/mariadb'), a path to a chart
 a packaged chart, or a fully qualified URL. For chart references, the latest
 version will be specified unless the '--version' flag is set.
 
+There are four different ways you can select the release name and namespace
+where the chart will be installed. By priority order:
```
viccuad, line 46:
```suggestion
where the chart will be upgraded. By priority order:
```

pkg/action/upgrade.go
```diff
@@ -57,3 +57,19 @@ func (i *Upgrade) Name(chart *chart.Chart, args []string) (string, error) {
 	// If we dont have our annotations then return the base name
 	return chart.Metadata.Name, nil
 }
+
+// SetNamespace sets the Namespace that should be used in action.Install
```
viccuad, line 61:
```suggestion
// SetNamespace sets the Namespace that should be used in action.Upgrade
```

cmd/hypper/upgrade.go
```diff
@@ -151,6 +165,13 @@ func newUpgradeCmd(cfg *action.Configuration, logger log.Logger) *cobra.Command
 				return err
 			}
 
+			// Set namespace for the upgrade client
+			if settings.NamespaceFromFlag {
+				client.Namespace = settings.Namespace()
+			} else {
+				client.SetNamespace(ch, settings.Namespace())
+			}
```
viccuad, line 173: Same here, aren't we missing setting the Config namespace with a call to `client.Config.SetNamespace(client.Namespace)`?
Analogous to: https://github.com/rancher-sandbox/hypper/blob/1a686aa3a18ebb10bbb2727baa28ea7e416ccc1a/cmd/hypper/install.go#L126

cmd/hypper/upgrade.go
```diff
@@ -114,6 +122,12 @@ func newUpgradeCmd(cfg *action.Configuration, logger log.Logger) *cobra.Command
 						logger.Info(eyecandy.ESPrintf(settings.NoEmojis, "Release %q does not exist. Installing it now. :toolbox:\n", client.ReleaseName))
 					}
 					instClient := action.NewInstall(cfg)
+					// Set namespace for the install client
+					if settings.NamespaceFromFlag {
+						instClient.Namespace = settings.Namespace()
+					} else {
+						instClient.SetNamespace(ch, settings.Namespace())
+					}
```
viccuad, line 130: Aren't we missing setting the Config namespace with a call to `client.Config.SetNamespace(client.Namespace)`?
Analogous to: https://github.com/rancher-sandbox/hypper/blob/1a686aa3a18ebb10bbb2727baa28ea7e416ccc1a/cmd/hypper/install.go#L126

viccuad (changes requested):

cmd/hypper/upgrade.go
Itxaka: yup yup!

cmd/hypper/upgrade.go
Itxaka: Agreed, card updated!

cmd/hypper/upgrade.go
Itxaka: :man_facepalming:

pkg/action/upgrade.go
Itxaka: :man_facepalming: :clown_face:

cmd/hypper/upgrade.go
dragonchaser: > I would prefer to see all duplications everywhere, to see the pattern, and then fix them in a way that is good for all callers, hence why I want to postpone that.

Well, why wait? I suggest using a helper package with static functions that can be invoked wherever needed... That can't hurt. If we see a pattern we can still move that function later. But we have it in place now.

cmd/hypper/upgrade.go
mattfarina: I think the lack of this line is causing the bug I found.

cmd/hypper/upgrade.go
Itxaka: I think this is out of scope for this, as this deals with a bug which needs to be fixed, not an improvement.

viccuad: I'm personally happy with the solution for now, we have the issue open to not forget about it.

pkg/action/upgrade.go
```diff
@@ -57,3 +57,19 @@ func (i *Upgrade) Name(chart *chart.Chart, args []string) (string, error) {
 	// If we dont have our annotations then return the base name
 	return chart.Metadata.Name, nil
 }
+
+// SetNamespace sets the Namespace that should be used in action.Upgrade
+//
+// This will read the chart annotations. If no annotations, it leave the existing ns in the action.
+func (i *Upgrade) SetNamespace(chart *chart.Chart, defaultns string) {
```
viccuad, line 64: We should add unit tests for this new function. E.g: https://github.com/rancher-sandbox/hypper/blob/viccuad%2Fadd-shared-deps/pkg/action/dependency_shared_test.go#L81-L101. The different TestSetNamespace() functions will clash, so either just rename the TestSetNamespace() to TestUpgradeSetNamespace(), or rename the others (as done here, but then I may need to rebase https://github.com/rancher-sandbox/hypper/pull/80/commits/a0acead55d686354d1a38ce5d9eb7128b8bbc46a#diff-b97a6f3723149f75e7945127b215944ea6a752bcf41a2e1c00b65ae0a409bc7cR34-R38).

viccuad (dismissed):

pkg/action/upgrade.go
Itxaka: good point, didnt even think about the name clash :o

Itxaka: unfortunately as I want to reuse some of your added functions I will have to wait for your Pr to merge :D

cmd/hypper/hypper_test.go
```diff
@@ -43,7 +43,7 @@ import (
 func testTimestamper() time.Time { return time.Unix(242085845, 0).UTC() }
 
 func init() {
-	action.Timestamper = testTimestamper
+	helmAction.Timestamper = testTimestamper
```
viccuad, line 46: :100:  top!

Itxaka:
Either this or it wazs already broken but passing the `release-name` flag does not work

Itxaka:
My mistake, seems that --release-name works, it was just missing tests...adding those...

pkg/action/upgrade.go
Itxaka: done!

cmd/hypper/upgrade.go
dragonchaser: Ok, then we should as this as new issue.

dragonchaser (approved):
lgtm

cmd/hypper/upgrade.go
Itxaka: got a PR already for this :D

Issue #75, opened 2021-03-09
install not setting the correct namespace if picked from annotations


Itxaka:
I want to:
 - Install the fleet[0] chart without specifying any annotations and let hypper install it to the proper namespace (fleet-system)
 - List all charts on the proper namespace and see the fleet chart

When I run `hypper install fleet fleet-0.3.000.tgz --create-namespace` I get:
```
Installing chart "fleet" in namespace "fleet-system"…
Done!
```

And I expect `hypper list` to not show the fleet chart as its not on the default namespace.
And I expect `hypper list --namespace fleet-system` to show the fleet chart

But I get:

```
$ bin/hypper list                           
NAME 	NAMESPACE   	REVISION	UPDATED                                	STATUS  	CHART        	APP VERSION
fleet	fleet-system	1       	2021-03-09 10:03:51.201245213 +0100 CET	deployed	fleet-0.3.000	0.3.0 

$ bin/hypper list --namespace fleet-system
NAME	NAMESPACE	REVISION	UPDATED	STATUS	CHART	APP VERSION
```

Some notes:
 - Chart is actually installed to proper namespace
 - Once installed with hypper, `helm list` shows the same exact output as hypper
 - If we set the `--namespace` flag to hypper then it works as expected
 - All resources are created in the proper namespace

Could this be a side issue of us managing the namespace and not setting it properly somewhere so when the chart is stored some fields are not properly stored? We may need @mattfarina expertise on Helm here to know where this could come from.

Itxaka:
The issue is that the storage client and the kubeclient have different namespaces. We set and modify the kubeclient one but the storage client is set automatically (can be modified via the --namespace flag only)

Storage client:
 - Stores the chart in a backend based on the driver(secrets, memory, sql, etc..) and segregated by the namespace
 - Namespace is set on https://github.com/helm/helm/blob/master/pkg/action/action.go#L365 when the file is loaded, and there is no external way of changing it. So when we first load the action file and the `Configuration` is init-ed, thats when the storage driver is initialized with the given namespace
 - Namespace loaded for the driver is the one in `EnvSettings.namespace` which is why it works in hypper when using the `---namespace` flag

Solutions?
 - Override the `Configuration.Init` function to set the namespace and then call the original one?? -> May even work, seems reasonable enough
 - Override the `EnvSettings` to manipulate the namespace there? -> Clumsy as hell, why there is chart loading code in envsettings, etc...doesnt look like a proper solution.
[...]

Issue #65, opened 2021-03-04
Use reflection to detect changes in the helm cli


dragonchaser:
When running the `hypper/cmd/helm`, check the composite code in `pkg/cli/` coming in from the helmcli type `EnvSettings` whether there have been fields added or not. If there are new fields do `os.Exit(1)` immediately. This is not a runtime feature but a feature for us to detect changes to the upstream (e.g. make the ci fail) code and react on it.

PR #76, opened 2021-03-09
Fix the disconnect between namespaces on install

Itxaka:
This re-Inits the config so we can set the Storage Driver
namespace to what is obtained from the chart.

This is because on helm you can only set the namespace via
the --namespace flag which gets into the envsettings.namespace
and the helm.go loads that var into the Configuration object
during cobra.OnInitialize.

To maintain compatibility we do the same but then during
the command execution we read the chart annotations and
change the release namespace on the fly.
This is ok for the namespace install but the Storage Driver,
which is already loaded, has a different namespace
(usually default) and the Release gets stored in the wrong
namespace (stored *for* helm, for k8s is installed properly)
This results into a disconnection between where the release
is installed and where hypper/helm looks in the Storage Driver
to search for it.

With this below we re-Init the configuration and hance the
Storage Driver to have the namespace correctly set in sync
with the Release

Maybe fixes: #75 ???

### DIFF
--- a/cmd/hypper/hypper.go
+++ b/cmd/hypper/hypper.go
@@ -62,7 +62,7 @@ func main() {
 	}
 
 	cobra.OnInitialize(func() {
-		helmDriver := os.Getenv("HELM_DRIVER")
+		helmDriver := settings.HelmDriver
 		if err := actionConfig.Init(settings.RESTClientGetter(), settings.Namespace(), helmDriver, logger.Debugf); err != nil {
 			log.Fatal(err)
 		}
@@ -127,5 +127,5 @@ func loadReleasesInMemory(actionConfig *action.Configuration) {
 // Storage Driver to search for it. With this below we re-Init the configuration and hance the Storage Driver to
 // have the namespace correctly set in sync with the Release
 func ReinitConfigForNamespace(cfg *action.Configuration, namespace string, logger log.Logger) {
-	_ = cfg.Init(settings.RESTClientGetter(), namespace, os.Getenv("HELM_DRIVER"), logger.Debugf)
+	_ = cfg.Init(settings.RESTClientGetter(), namespace, settings.HelmDriver, logger.Debugf)
 }
--- a/cmd/hypper/hypper.go
+++ b/cmd/hypper/hypper.go
@@ -114,3 +114,18 @@ func loadReleasesInMemory(actionConfig *action.Configuration) {
 	// Must reset namespace to the proper one
 	mem.SetNamespace(settings.Namespace())
 }
+
+// This re-Inits the config so we can set the Storage Driver namespace to what is obtained
+// from the chart.
+// This is because on helm you can only set the namespace via the --namespace flag which gets into
+// the envsettings.namespace and the helm.go loads that var into the Configuration object during
+// cobra.OnInitialize. To maintain compativility we do the same but then during the command execution
+// we read the chart annotations and change the release namespace on the fly. This is ok for the
+// namespace install but the Storage Driver, which is already loaded, has a different namespace (usually default)
+// and the Release gets stored in the wrong namespace (stored *for* helm, for k8s is installed properly)
+// This results into a disconnect between where the release is installed and where hypper/helm looks in the
+// Storage Driver to search for it. With this below we re-Init the configuration and hance the Storage Driver to
+// have the namespace correctly set in sync with the Release
+func ReinitConfigForNamespace(cfg *action.Configuration, namespace string, logger log.Logger) {
+	_ = cfg.Init(settings.RESTClientGetter(), namespace, os.Getenv("HELM_DRIVER"), logger.Debugf)
+}
--- a/cmd/hypper/install.go
+++ b/cmd/hypper/install.go
@@ -17,7 +17,11 @@ limitations under the License.
 package main
 
 import (
+	"github.com/Masterminds/log-go"
+	logio "github.com/Masterminds/log-go/io"
 	"github.com/pkg/errors"
+	"github.com/rancher-sandbox/hypper/pkg/action"
+	"github.com/rancher-sandbox/hypper/pkg/eyecandy"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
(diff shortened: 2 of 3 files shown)


Itxaka:
test are failing, I know, first I want to see if this is a viable solution. I do not understand the full implications of re-Init the config. Locally it seems ok, I could not find anything that was missed by re-initializing but I rather have @mattfarina have a good look and maybe offer an alternative if its simpler.


BTW some tests seems to be fail now like ....properly. Not able to install to X namespace becuase the namespace doesnt exists seems like the proper outcome, so this may have fixed the issue actually and the tests were just installing into the default namespace!

also some upgrade test which I probably broke or was broken since day one :clown_face:

cmd/hypper/hypper.go
```diff
@@ -114,3 +114,18 @@ func loadReleasesInMemory(actionConfig *action.Configuration) {
...
+// the envsettings.namespace and the helm.go loads that var into the Configuration object during
+// cobra.OnInitialize. To maintain compativility we do the same but then during the command execution
+// we read the chart annotations and change the release namespace on the fly. This is ok for the
+// namespace install but the Storage Driver, which is already loaded, has a different namespace (usually default)
+// and the Release gets stored in the wrong namespace (stored *for* helm, for k8s is installed properly)
+// This results into a disconnect between where the release is installed and where hypper/helm looks in the
+// Storage Driver to search for it. With this below we re-Init the configuration and hance the Storage Driver to
+// have the namespace correctly set in sync with the Release
+func ReinitConfigForNamespace(cfg *action.Configuration, namespace string, logger log.Logger) {
+	_ = cfg.Init(settings.RESTClientGetter(), namespace, os.Getenv("HELM_DRIVER"), logger.Debugf)
```
dragonchaser, line 130: Reading env variables outside environment.go... I do not really like that, can we move that to a settings variable?

Just saw this is also down when assigning the `helmDriver` variable in line 65 and in `cmd/hypper/list.go`.....

dragonchaser (changes requested):

cmd/hypper/hypper.go
Itxaka: well, I was trying to mimick the first `Init` which is done just above in line 65. This is how helm does it, so Im guessing we just lift it from there?

Itxaka: I wonder why they dont directly use `settings` here to set both of those options...maybe an oversight??

Itxaka: updated to use settings as it turns out that we can add it there with no issues that I can see. I wonder if we should upstream this and the other env var used...

cmd/hypper/install.go
```diff
@@ -120,6 +119,9 @@ func runInstall(args []string, client *action.Install, valueOpts *values.Options
 		client.SetNamespace(chartRequested, settings.Namespace())
 	}
 
+	// Reinit the config so we set the proper namespace of the Storage Driver
+	ReinitConfigForNamespace(client.Config, client.Namespace, logger)
```
viccuad, line 123: Why not directly do `client.Config.Init(settings.RESTClientGetter(), client.Namespace, settings.HelmDriver, logger.Debugf)`?
I find it more readable that a static function that contains no logic, and no changes in the scope of vars/functions it access.

cmd/hypper/hypper.go
```diff
@@ -114,3 +114,18 @@ func loadReleasesInMemory(actionConfig *action.Configuration) {
 	// Must reset namespace to the proper one
 	mem.SetNamespace(settings.Namespace())
 }
+
+// This re-Inits the config so we can set the Storage Driver namespace to what is obtained
```
viccuad, line 118: This comment is super useful, thanks! :D

nit: But I think it's maybe to big. I would prefer if it were on the commit message of the commit that adds the fix, accessible from git blame.

pkg/cli/environment.go
```diff
@@ -95,6 +96,7 @@ func New() *EnvSettings {
 		RegistryConfig:   envOr("HYPPER_REGISTRY_CONFIG", hypperpath.ConfigPath("registry.json")),
 		RepositoryConfig: envOr("HYPPER_REPOSITORY_CONFIG", hypperpath.ConfigPath("repositories.yaml")),
 		RepositoryCache:  envOr("HYPPER_REPOSITORY_CACHE", hypperpath.CachePath("repository")),
+		HelmDriver:       os.Getenv("HELM_DRIVER"),
```
viccuad, line 99: :100: !

viccuad (changes requested):
Wooho! good find! :D
And the helmDriver change is also a win!

cmd/hypper/install.go
viccuad: Shouldn't we have access to client.Config.Init there, too?

cmd/hypper/hypper.go
viccuad: I like the one-line comment `// Reinit the config so we set the proper namespace of the Storage Driver` when it's called, with that is obvious that the call has a purpose. But I'm ok either way, it's a nit :).

cmd/hypper/install.go
Itxaka: we should but if we are gonna reuse this in teh upgrade for example I rather have a common function that sets this.

In any case this does not work as expected as it clears the full KubeClient ¬_¬

Itxaka:
Unfortunately, not a good solution as we wipe out a lot of stuff :(

pkg/cli/environment.go
viccuad: ok, I don't like changing `HELM_DRIVER` to `HYPPER_DRIVER` anymore. If we change it, we need to also start overloading `ActionConfig.Init()` and wherever it gets used as `HELM_DRIVER`, to `HYPPER_DRIVER`. Example:

https://github.com/helm/helm/blob/6e4a61f21279560b4a4cf7b84088c9e7e75757a9/pkg/action/action.go#L411

dragonchaser: Yes but this calls of implementing #65 urgently so we do not miss any changes there....

PR #79, opened 2021-03-09
fix(storage): Fixing namespace of install storage

### LINKED ISSUE #75: install not setting the correct namespace if picked from annotations

mattfarina:
Closes #75

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -23,6 +23,8 @@ require (
 	golang.org/x/term v0.0.0-20210220032956-6a3ed077a48d
 	gopkg.in/yaml.v2 v2.4.0
 	helm.sh/helm/v3 v3.5.2
+	k8s.io/api v0.20.4
+	k8s.io/apimachinery v0.20.4
 	k8s.io/cli-runtime v0.20.4
 	k8s.io/client-go v0.20.4
 	sigs.k8s.io/yaml v1.2.0
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -16,8 +16,13 @@ limitations under the License.
 package action
 
 import (
+	"fmt"
+	"os"
+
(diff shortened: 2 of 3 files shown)


Itxaka (approved):

viccuad (approved):
Nice!
`lazyclient.go` contents can't be easily composed, as the type is unexported. But I would prefer to have its copy-paste inclusion in its own commit, with the commit message specifying at what point from Helm's history it was taken from.

Approving anyways, to have some speed.

Issue #30, opened 2021-01-29
Shared Dependencies


mattfarina:
Those who setup Helm repos can have a collection of Helm charts that install and work together. But, they don't all need to be installed at the same time or together.

For example, there are two apps that independent of each other. But, you want both of them to work with a shared prometheus. In Helm, you would need to install prometheus. Then when installing the application know where prometheus is and pass in the configuration details.

This setup would enable a chart to specify the name and namespace to be installed into. This would be a default value to be used. Then other charts could specify the dependency they need, the name to look for, and the namespace to look in. If the dependency does not exist it should then be installed first. If it is installed than it can be used.

RBAC needs to be follows and if access is not available hypper should fail and notify users.

This issue only needs to deal with the installation of the shared dependencies. Configuring them to work together is a separate concern.

Requiring a shared dependency can be captured in Helm chart annotations like:

```yaml
annotations:
  hypper.cattle.io/shared-dependencies: |
   - name: bar            # Chart instance name in the namespace
     namespace: default   # The namespace to look in
     chart: foo           # The name of the chart
     version: ^1.2.3      # A version range of supported versions
```

Installing to a common location is covered in #29.

mattfarina:
See https://github.com/rancher-sandbox/hypper/blob/main/docs/shared-dependencies.md for details on the design.

kkaempf:
Thinking about it, why is it called 'shared' dependencies ?

In my understanding, we're talking about strong/weak dependencies only. The 'shared' part is actually an admin choice of re-using an already deployed service.

mattfarina:
@kkaempf Ultimately, there are 3 types of dependencies we will setup.

1. Required dependencies that are to be installed uniquely. For example, you have a wordpress chart that requires mysql. If wordpress is installed 3 times than you will have 3 mysql installs. Each of the mysql installs is tightly coupled to its corresponding wordpress install. Helm provides this today. We do not need to implement it.
2. Required dependencies that are shared. So wordpress requires mysql. If the wordpress chart is installed 3 times there is just one mysql that is shared. wordpress and mysql are a bad example but allow to illustrate the idea. The dependencies are shared which is different from the default way provided by Helm. Hypper is tightly coupled to Helm so we need to explain this difference and note it.
3. Optional dependencies. We'll get to this in a little bit.

Do you have a suggestion for another name for shared dependencies?

viccuad:
Dropping myself from the card from now, I'm still focusing on the repo card.
[...]

PR #80, opened 2021-03-10
feat: Add `hypper shared-dep list` command

viccuad:
This PR Implements `hypper shared-dep list` command:
   ```
   $ hypper shared-dep list ~/suse/hypper/cmd/hypper/testdata/testcharts/hypper-annot
   NAME            VERSION REPOSITORY                              STATUS
   prometheus      13.3.1  https://example.com/charts              deployed
   postgresql      10.3.11 https://another.example.com/charts      not-installed
   ```
It does that by:
- Creating `pkg/action/dependency_shared{,_test}.go` and tests
- Creating `cmd/hypper/dependency_shared{,_test}.go` and tests

Partially implements https://github.com/rancher-sandbox/hypper/issues/30 (missing installing shared-deps when doing `hypper install`).

### DIFF
--- a/cmd/hypper/dependency_shared.go
+++ b/cmd/hypper/dependency_shared.go
@@ -1,0 +1,119 @@
+/*
+Copyright The Helm Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"io"
+	"path/filepath"
+
+	"github.com/spf13/cobra"
+
+	"helm.sh/helm/v3/cmd/helm/require"
+	"helm.sh/helm/v3/pkg/action"
+)
+
+const dependencyDesc = `
+Manage the dependencies of a chart.
+
+Helm charts store their dependencies in 'charts/'. For chart developers, it is
+often easier to manage dependencies in 'Chart.yaml' which declares all
+dependencies.
+
+The dependency commands operate on that file, making it easy to synchronize
+between the desired dependencies and the actual dependencies stored in the
+'charts/' directory.
+
+For example, this Chart.yaml declares two dependencies:
+
+    # Chart.yaml
+    dependencies:
+    - name: nginx
+      version: "1.2.3"
+      repository: "https://example.com/charts"
+    - name: memcached
+      version: "3.2.1"
+      repository: "https://another.example.com/charts"
+
+
+The 'name' should be the name of a chart, where that name must match the name
+in that chart's 'Chart.yaml' file.
+
+The 'version' field should contain a semantic version or version range.
+
+The 'repository' URL should point to a Chart Repository. Helm expects that by
+appending '/index.yaml' to the URL, it should be able to retrieve the chart
+repository's index. Note: 'repository' can be an alias. The alias must start
+with 'alias:' or '@'.
+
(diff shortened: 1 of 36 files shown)


viccuad:
Failing because of https://github.com/rancher-sandbox/hypper/pull/80/commits/5349efb70cee55edcb5350f9496cbdf06fc6f0b2, as the stores are expecting a local cluster..

cmd/hypper/testdata/testcharts/hypper-annot/Chart.yaml
```diff
@@ -10,3 +10,10 @@ annotations:
   hypper.cattle.io/release-name: my-hypper-name
   catalog.cattle.io/namespace: fleet-system
   catalog.cattle.io/release-name: fleet
+  hypper.cattle.io/shared-dependencies: |
```
Itxaka, line 13: can we create a different chart for this so we can manipulate tests without affecting other tests?

cmd/hypper/dependency_shared.go
```diff
@@ -0,0 +1,133 @@
...
+func newSharedDependencyCmd(cfg *action.Configuration, logger log.Logger) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "shared-dep list",
+		Aliases: []string{"shared-deps", "shared-dependencies"},
+		Short:   "manage a chart's shared dependencies",
+		Long:    sharedDependencyDesc,
+		Args:    require.NoArgs,
+	}
+
+	cmd.AddCommand(newSharedDependencyListCmd(cfg, logger))
```
Itxaka, line 92: Im not understanding why we have 2 `cobra.Command` here. Is this a subcommand of `shared-dep list` somehow? Really not understanding as I would expect just one command `shared-dep` and that having subcommands then (list, show, whatever) or its just the `Use` value wrong and it should be `shared-dep SUBCOMMAND` ?

Itxaka (commented):
lookign good, a couple of questions

cmd/hypper/dependency_shared.go
viccuad: Yep, we have the command `shared-dep` and then the subcommand `shared-dep list` (analogous to Helm's `dependency build | list | update`). I went with the command + subcommand because it allows us to have the help info, if follows the same UI as Helm (`s/depedency/shared-dep`), and because it really feels that when installing shared-deps, we will add subcommands for installing/updating shared-deps, apart from doing it on `hypper install`, or getting a graph of shared-deps.

The `Use` in the future could look like `shared-dep install | list | update`. Now, it only has `list`.

cmd/hypper/testdata/testcharts/hypper-annot/Chart.yaml
viccuad: Mmh, I'm not sold on that. Thing is, by adding this annotations to that chart, we just made sure that the previous tests, which pass unchanged, keep working with the `shared-dependencies` annotation. That way we don't need to create additional tests for now. Is that convincing enough? :>

In the future we may need additional testcharts for testing install of chart+shared-deps, though.

Itxaka: its just that if we now extend the install tests for example and modify this chart (I think we use it there) we could potentially be breaking tests for shared-deps by manipulating it.

In any case we can cross that bridge when we reach it :)

cmd/hypper/dependency_shared.go
Itxaka: ahh, makes sense! Thanks!

viccuad:
Ready for review! Initialized correctly the testing release stores with https://github.com/rancher-sandbox/hypper/pull/80/commits/c4169878dc25ff247428c7d866525ec1f084b8e9. Also, lots of force-pushes to fill the testdata with windows output, sorry for the noise. Might be time to develop in a personal fork..

cmd/hypper/dependency_shared.go
viccuad: resolving as I suppose it's ok! :)

Itxaka (approved):
Looks good to me!

mattfarina (approved):

Issue #62, opened 2021-03-03
Deterministic release name generation


mattfarina:
In the current setup the release name is either specified, pulled from the chart metadata, or generated when a flag is passed in. Darren suggested automatically generating a release name in a deterministic manner when a release name is not specified.

dragonchaser:
What exactly do you have in mind? Like one of those random-three-noun name generators jitsi uses or the chart file name appended with a random n-digit numeral?

kkaempf:
This looks easy and relieves users from naming things. Adding mvp label :wink:

viccuad:
Right now when one does `hypper install --generate-name`, we create a name as `chart + current time`, just like Helm does (see https://github.com/rancher-sandbox/hypper/blob/main/pkg/action/install.go#L126-L139).

We could create a name as `chart + uri/repo`, `chart + sha(uri/repo)`, or something similar. Or use the `sha(uri + chartname)` to then obtain one of those cute 2-words names as docker does.

mattfarina:
Helm v2 used to do something similar to moby by leveraging https://github.com/technosophos/moniker. If I understand the moby code right, it's random rather than deterministic. The ask is for a deterministic rather than random solution. That way I could install the chart 5 times and get the same name every time.

dragonchaser:
Right, so if I understand it right we have to generate the names once (for each chart) and have to store it somewhere. Next question is, should it be deterministic per hypper client (would result us in storing those names in a map in the hypper cache) or should it be deterministic and identical for every cluster admin (would result in implementing a storage for that on the cluster).

viccuad:
if it is deterministic, one doesn't need to save the release name. To obtain a release name that corresponds to that chart (+ repo url), one just calls the same function to compute the name.
If we add cluster information to the name creation, then, the names will be only deterministic inside that cluster, but there wouldn't be any need to save the name in a different cache or db.

dragonchaser:
I am not sure if I am following you here, can you give me an example for what the name would look like for a chart named `my-awesome-chart.tar.gz` for example?

viccuad:
let's say that you are installing `my-awesome-chart.tar.gz` from the chartpath `~/downloads/my-local-repo`. The name would be something computed from `name(my-awesome-chart, hash512sum(~/downloads/my-local-repo`))`, returning something like `my-awesome-chart-aefd52365g`.

[...]

PR #81, opened 2021-03-10
Deterministic names

dragonchaser:
This PR removes the `--generate-name` flag, if now chart name is provided (either through annotations or cli) it will use the name from the `Chart.yaml` out of the chart itself.

### DIFF
--- a/pkg/action/install_test.go
+++ b/pkg/action/install_test.go
@@ -106,11 +106,11 @@ func TestName(t *testing.T) {
 	instAction = installAction(t)
 	instAction.ReleaseName = ""
 	chart = buildChart()
-	_, err = instAction.Name(chart, []string{"chart-uri"})
+	name, err = instAction.Name(chart, []string{"chart-uri"})
 	if err != nil {
 		t.Fatal(err)
 	}
-	is.Equal("fleet", name)
+	is.Equal("hello", name)
 }
 
 func TestChart(t *testing.T) {
--- a/pkg/action/install_test.go
+++ b/pkg/action/install_test.go
@@ -119,10 +119,10 @@ func TestName(t *testing.T) {
 	instAction.ReleaseName = ""
 	chart = buildChart()
 	_, err = instAction.Name(chart, []string{"chart-uri"})
-	if err == nil {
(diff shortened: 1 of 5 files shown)


dragonchaser:
Blocked, see discussion #62

dragonchaser:
We decided in the daily to drop the `--generate-name` flag.

pkg/action/install_test.go
```diff
@@ -104,85 +102,15 @@ func TestName(t *testing.T) {
...
-	if err == nil {
-		t.Fatal(err)
-	}
-	is.Equal("cannot set --generate-name and also specify a name", err.Error())
-
 	// no name or annotations present
 	instAction = installAction(t)
 	instAction.ReleaseName = ""
 	chart = buildChart()
 	_, err = instAction.Name(chart, []string{"chart-uri"})
```
viccuad, line 109:
```suggestion
	name, err = instAction.Name(chart, []string{"chart-uri"})
```
If not, we will be doing `is.Equal()` of the previous `name` value.

pkg/action/install_test.go
```diff
@@ -104,85 +102,15 @@ func TestName(t *testing.T) {
...
-			if err != nil {
-				t.Fatal(err)
-			}
-
-			is.Equal(tc.ExpectedName, name)
-		})
+	if err != nil {
+		t.Fatal(err)
 	}
+	is.Equal("fleet", name)
```
viccuad, line 113:
```suggestion
	is.Equal("hello", name)
```

It should be hello, the chart was created with `buildchart()`, not with `buildchart(withFallbackAnnot())`. Together with the other change, makes everything peachy :).

viccuad (commented):
Looks good! But the tests need to be fixed so the correct test is being carried out.

Itxaka:
Agreed with Victor, code seems good but good catch on the tests!

pkg/action/install_test.go
dragonchaser: That would explain the bug in Line 113....

Itxaka (approved):
looks good!

Issue #77, opened 2021-03-09
`hypper uninstall` not working correctly


kkaempf:
This is probably related to #73 and #75.

## tl;dr

`hypper uninstall` fails where `helm uninstall` succeeds.

---

I'm following the [chimera-controller install instructions](https://github.com/chimera-kube/chimera-controller#installation), replacing `helm` with `hypper`:

```
$ hypper install --namespace chimera --create-namespace chimera-controller chimera/chimera-controller
```

A subsequent uninstall, although reported as successfull, doesn't actually uninstall the controller.

```
$ hypper uninstall --namespace chimera chimera-controller
🔥  uninstalling chimera-controller
🔥  release "chimera-controller" uninstalled
```

`k9s` still shows `chimera-controller` as running.

Repeating the `hypper uninstall` results in:
```
hypper uninstall --namespace chimera chimera-controller
🔥  uninstalling chimera-controller
Error: uninstall: Release not loaded: chimera-controller: release: not found
```

Doing the same install/uninstall sequence with `helm` works as expected.

dragonchaser:
Is it possible that this is the same issue like #74 ?

viccuad:
Just tried on current main, and the issue is still there :crying_cat_face: .

Itxaka:
probably the same issue as  https://github.com/rancher-sandbox/hypper/issues/75 but we broke the `--namespace` flag?

Itxaka:
should be fixed on main! Keep the bugs coming @kkaempf !

kkaempf:
Reopening - still not fixed.

Note: I renamed `helm` to `helm3` and `hypper` to `helm`.

### How to reproduce

- install [Carrier](https://github.com/SUSE/carrier) on [minikube](https://github.com/SUSE/carrier/blob/main/docs/install.md#minikube)
- run `carrier install`
- run `carrier uninstall`

The above fails (with hypper) at
> helm uninstall service-catalog --namespace service-catalog

with a timeout.

Running with the real helm succeeds.

kkaempf:
It's _only_ the 'service-catalog' step that fail, all other `carrier uninstall` commands (calling hypper as helm) succeed.

kkaempf:
Carrier's code for service-catalog management: https://github.com/SUSE/carrier/blob/main/deployments/service_catalog.go

Itxaka:
Writing it down so I dont forget.

[...]

PR #82, opened 2021-03-11
Set the client namespace on uninstall

Itxaka:
We are missing reading from the chart annotations as well,
or maybe thats not gonna be implemented?

We are missing this piece to be able to clean up the resources properly but probably missing obtaining the namespace from annotattions which is quite bigger....and would probably benefit from extraction some generic methods...

Partial fix: #77 

### DIFF
--- a/cmd/hypper/uninstall.go
+++ b/cmd/hypper/uninstall.go
@@ -40,6 +40,7 @@ func newUninstallCmd(actionConfig *action.Configuration, logger log.Logger) *cob
 		RunE: func(cmd *cobra.Command, args []string) error {
 			for i := 0; i < len(args); i++ {
 				logger.Info(eyecandy.ESPrintf(settings.NoEmojis, ":fire: uninstalling %s", args[i]))
+				client.Config.SetNamespace(settings.Namespace())
 				res, err := client.Run(args[i])
 				if err != nil {
 					return err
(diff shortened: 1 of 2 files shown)


viccuad (approved):

viccuad:
Looks good! I think this is it. And I don't think there's any need to look for annotations; the `uninstall` command works on releases on the store, not on charts, so there's no annotations to be looked at!

viccuad:
I think we should add some tests to catch the problems (maybe out of this PR). Adding tests to `cmd/hypper/uninstall_test.go` doesn't give us anything, as those check for command output. I thought adding `pkg/action/uninstall_test.go` shouldn't be needed, but as we are using our own `Config` instead of the one inside of helm/pkg/action/install.go, then we should write tests testing our `Config`.

Itxaka:
> I think we should add some tests to catch the problems (maybe out of this PR). Adding tests to cmd/hypper/uninstall_test.go doesn't give us anything, as those check for command output. I thought adding pkg/action/uninstall_test.go shouldn't be needed, but as we are using our own Config instead of the one inside of helm/pkg/action/install.go, then we should write tests testing our Config.

Do we really? Like, are we doing something any special that setNamespace (which should be tested of course)? The rest is basically the same as helm and should be covered by helm testing on their side no?

Issue #61, opened 2021-03-03
Create User Docs - Tutorials/get-started


mattfarina:
Following https://documentation.divio.com/tutorials/, add Hypper tutorials that tells end-users how to:

- [ ] Install Hypper
- [x] Use Hypper to install, upgrade, and otherwise work with charts
...

Itxaka:
Should we:
 - Copy Helm docs and expand on those with Hypper specifics
 or
 - Write specific Hypper docs and point to Helm docs for the generic stuff

kkaempf:
imho, Hypper should stay an extension on top of Helm and be a full replacement. Cluster operators shouldn't need to think about "do I need Hypper for this or Helm ?" but always use Hypper. That said, I'd prefer the "write specific Hypper docs" option.

viccuad:
Edited the issue and split it into "Tutorials" and "Explanations", and added issues for "How-tos" and "Reference guide".

Itxaka:
Part of this was already done by @viccuad on https://github.com/rancher-sandbox/hypper/pull/86

So only the install part is needed now :)

PR #86, opened 2021-03-12
Add docs/user/tutorials/{installing,quickstart}.md

viccuad:
Based on https://github.com/helm/helm-www/tree/master/content/en/docs/intro.

Partially implements https://github.com/rancher-sandbox/hypper/issues/61.

### DIFF
--- a/docs/user/tutorials/installing.md
+++ b/docs/user/tutorials/installing.md
@@ -1,0 +1,27 @@
+# Installing Hypper
+
+## From source (Linux, macOS)
+
+Building Hypper from source is a bit more work, but the best way to test the
+latest (pre-release) Hypper version.
+
+You must have a working Go environment (1.14+).
+
+```terminal
+$ git clone https://github.com/rancher-sandbox/hypper/hypper.git
+$ cd hypper
+$ make
+```
+
+If required, it will fetch the dependencies and cache them, and validate
+configuration.
+It will then compile `hypper`, and place it in `./bin/hypper`.
+
+
+## From development builds
+
+TBD.
+
+## From Github releases
+
+TBD.
--- a/docs/user/tutorials/installing.md
+++ b/docs/user/tutorials/installing.md
@@ -20,8 +20,12 @@ It will then compile `hypper`, and place it in `./bin/hypper`.
 
 ## From development builds
 
-TBD.
+In addition to releases you can download and install development snapshots of
+Hypper from [Github Actions].
 
 ## From Github releases
 
 TBD.
+
+
+[Github Actions]: https://github.com/rancher-sandbox/hypper/actions/workflows/ci.yml?query=branch%3Amain+is%3Asuccess+workflow%3ACI
--- a/docs/user/tutorials/quickstart.md
+++ b/docs/user/tutorials/quickstart.md
@@ -62,8 +62,8 @@ this case the official stable charts.
 
 ```terminal
 $ hypper repo update              # Make sure we get the latest list of charts
-$ hypper install bitnami/mariadb --generate-name
-Installing chart "mariadb-1615547386" in namespace "default"…
+$ hypper install mariadb bitnami/mariadb
+Installing chart "mariadb" in namespace "default"…
 Done! 👏
 ```
 
@@ -101,9 +101,9 @@ It's easy to see what has been released with hypper:
 
 ```terminal
 $ hypper ls --all-namespaces
-NAME                    NAMESPACE       REVISION        UPDATED                                 STATUS          CHART                           APP VERSION
-fleet                   fleet-system    1               2021-03-12 12:06:35.951012048 +0100 CET deployed        fleet-0.3.400                   0.3.4
-mariadb-1615547386      default         1               2021-03-12 12:09:46.670491535 +0100 CET deployed        mariadb-9.3.5                   10.5.9
+NAME       NAMESPACE       REVISION        UPDATED                                 STATUS          CHART                 APP VERSION
+fleet      fleet-system    1               2021-03-12 12:06:35.951012048 +0100 CET deployed        fleet-0.3.400         0.3.4
+mariadb    default         1               2021-03-12 12:09:46.670491535 +0100 CET deployed        mariadb-9.3.5         10.5.9
 ```
 
 ## Uninstall a release
--- a/docs/user/tutorials/quickstart.md
+++ b/docs/user/tutorials/quickstart.md
@@ -2,30 +2,20 @@
 
 This guide covers how you can quickly get started using Hypper.
 
-If you already know Helm, this will be a breeze, as Hypper follows Helm's UI,
-being a drop-in replacement, and extends its functionalities to install Helm
-and Hypper charts alike.
+If you already know Helm, this will be a breeze, as Hypper follows Helm's workflow
+and extends its functionalities to install Helm and charts that have been extended
+to work with Hypper.
 
 ## Prerequisites
 
-The following prerequisites are required for a successful and properly secured
-use of Hypper.
+The following prerequisites are required:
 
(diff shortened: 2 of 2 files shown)


docs/user/tutorials/quickstart.md
```diff
@@ -0,0 +1,138 @@
...
+and Hypper charts alike.
+
+## Prerequisites
+
+The following prerequisites are required for a successful and properly secured
+use of Hypper.
+
+1. A Kubernetes cluster.
+1. Deciding what security configurations to apply to your installation, if any
+1. Installing and configuring Hypper.
```
mattfarina, line 16: I don't think we need to talk about securing a cluster. People may just kick the tires in something like Docker Desktop. All they need is a Kubernetes cluster, Hypper, and a chart to install.

docs/user/tutorials/quickstart.md
```diff
@@ -0,0 +1,138 @@
...
+
+
+## Install an example Helm chart
+
+To install a chart, you can run the hypper install command. Hypper has several
+ways to find and install a chart, but the easiest is to use a repository, in
+this case the official stable charts.
+
+```terminal
+$ hypper repo update              # Make sure we get the latest list of charts
```
mattfarina, line 64: If a user just ran `hypper repo add` wouldn't they already have the up to date charts. Is this needed?

docs/user/tutorials/quickstart.md
```diff
@@ -0,0 +1,138 @@
...
+$ hypper install mariadb bitnami/mariadb
+Installing chart "mariadb" in namespace "default"…
+Done! 👏
+```
+
+As with Helm, you can get an idea of the features of this MariaDB chart by
+running `hypper show chart bitnami/mariadb`, or `hypper show all
+bitnami/mariadb` to get all information about the chart.
```
mattfarina, line 72: We've not implemented `hypper show` yet

docs/user/tutorials/quickstart.md
```diff
@@ -0,0 +1,138 @@
...
+
+Hypper enables you to deploy Helm and Hypper charts into your Kubernetes
+cluster. To simplify that, the best way is to add a chart repository, for
+example one with Helm charts:
+
+```terminal
+$ hypper repo add bitnami https://charts.bitnami.com/bitnami
+```
+
+You can add several repositories. Let's add also a repository containig charts
```
Itxaka, line 40: typo

docs/user/tutorials/quickstart.md
```diff
@@ -0,0 +1,138 @@
...
+The commands are the same as you have already used:
+
+```terminal
+$ hypper install rancher-charts/fleet --create-namespace
+Installing chart "fleet" in namespace "fleet-system"…
+Done! 👏
+```
+
+This time, the chart got installed with a default name `fleet`, and into a
+default namespace  `fleet-system`. We passed the flag `--create-namespace` to
```
Itxaka, line 95: `to not need to create` sounds a bit off methinks. `so we dont need to create` maybe?

docs/user/tutorials/quickstart.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Install an example Hypper chart
+
+Hypper charts are supersets of Helm charts with more functionality, such as
+specifying default release name, namespace of installation, or installing
+shared-dependency charts automatically.
+They are thought to be installed system-wide in the cluster: 1 Hypper chart for
+all users of the cluster. Think of them as typical system OS libraries/services.
+
+The commands are the same as you have already used:
```
Itxaka, line 86: Im not sure if this is referring to a previous step on the same tutorial or it references that if you have used Helm before you are used to this commands. Maybe we should eb more explicit here?

docs/user/tutorials/quickstart.md
Itxaka: `let's also ad` instead of `let's add also` ...or maybe Im wrong? Im not sure which one is the correct one..maybe both are?

docs/user/tutorials/quickstart.md
```diff
@@ -0,0 +1,138 @@
...
+```
+
+You can add several repositories. Let's add also a repository containig charts
+with some Hypper functionality:
+
+```terminal
+$ hypper repo add rancher-charts https://charts.rancher.io
+```
+
+Now, you can list the repository that Hypper can access:
```
Itxaka, line 47: should we mention `that hypper can install charts from` instead to make it clearer?

docs/user/tutorials/quickstart.md
```diff
@@ -0,0 +1,138 @@
...
+
+The following prerequisites are required for a successful and properly secured
+use of Hypper.
+
+1. A Kubernetes cluster.
+1. Deciding what security configurations to apply to your installation, if any
+1. Installing and configuring Hypper.
+
+
+## Install Kubernetes or have access to a cluster
```
Itxaka, line 19: should we link to the official k8s docs here?

docs/user/tutorials/installing.md
```diff
@@ -0,0 +1,31 @@
...
+configuration.
+It will then compile `hypper`, and place it in `./bin/hypper`.
+
+
+## From development builds
+
+In addition to releases you can download and install development snapshots of
+Hypper from [Github Actions].
+
+## From Github releases
```
dragonchaser, line 26: As soon as this is decided, this should be the first item in this document.

docs/user/tutorials/quickstart.md
```diff
@@ -0,0 +1,138 @@
+# Quickstart
+
+This guide covers how you can quickly get started using Hypper.
+
+If you already know Helm, this will be a breeze, as Hypper follows Helm's UI,
```
dragonchaser, line 5: I do not like the term UI here, I'd opt for something like `... as Hypper follows Helm's workflow,...`

docs/user/tutorials/quickstart.md
dragonchaser: Either that or rancher docs....

docs/user/tutorials/quickstart.md
dragonchaser: In that case it should not bee needed.

dragonchaser (changes requested):

docs/user/tutorials/installing.md
mattfarina: I would not be surprised if we have a get script to handle the install. We want to make it easy. In any case, this section will get rewritten at least once.

docs/user/tutorials/quickstart.md
mattfarina: There are so many ways to get a Kubernetes cluster. We shouldn't be telling people how to do it. People who come to check this out will already have a way to get Kubernetes.

mattfarina (approved):
This is a good start. The feedback can be incorporated in a follow-up PR since Victor isn't available to make changes.

docs/user/tutorials/quickstart.md
mattfarina: I reworked this paragraph as Hypper is no longer a drop-in replacement.

docs/user/tutorials/quickstart.md
mattfarina: Good catches.

Itxaka (approved):
Good enough for a start!

docs/user/tutorials/quickstart.md
dragonchaser: I think this is sufficient for a quick start, everybody familiar with helm will assume that these commands work from the first paragraph of this document, this should IMHO be more helpful to as an entry-point to the inexperienced user.

dragonchaser (approved):

PR #89, opened 2021-03-15
Introduce common methods for setNamespace and Name

Itxaka:
As both install and upgrade are using the action.SetNamespace
method, this extracts that method into a different file and
decouples it from the action so its easier to share and reuse

At the same time expands the method so it checks for the
NamespaceFromFlag and sets the client.Config storage
driver to the same namespace as the client in one go,
so we dont have to call multiple places for a namespace change

### DIFF
--- a/cmd/hypper/install.go
+++ b/cmd/hypper/install.go
@@ -116,13 +116,8 @@ func runInstall(args []string, client *action.Install, valueOpts *values.Options
 		return nil, err
 	}
 
-	if settings.NamespaceFromFlag {
-		client.Namespace = settings.Namespace()
-	} else {
-		client.SetNamespace(chartRequested, settings.Namespace())
-	}
-
-	client.Config.SetNamespace(client.Namespace)
+	// Set namespace for the install client
+	action.SetNamespace(client, chartRequested, settings.Namespace(), settings.NamespaceFromFlag)
 
 	name, err := client.Name(chartRequested, args)
 	if err != nil {
--- a/cmd/hypper/install.go
+++ b/cmd/hypper/install.go
@@ -119,11 +119,12 @@ func runInstall(args []string, client *action.Install, valueOpts *values.Options
 	// Set namespace for the install client
 	action.SetNamespace(client, chartRequested, settings.Namespace(), settings.NamespaceFromFlag)
 
-	name, err := client.Name(chartRequested, args)
-	if err != nil {
-		return nil, err
+	if client.ReleaseName == "" {
+		client.ReleaseName, err = action.GetName(chartRequested, client.NameTemplate, args...)
+		if err != nil {
+			return nil, err
+		}
 	}
-	client.ReleaseName = name
 
 	if err := checkIfInstallable(chartRequested); err != nil {
 		return nil, err
--- a/cmd/hypper/upgrade.go
+++ b/cmd/hypper/upgrade.go
@@ -106,7 +106,10 @@ func newUpgradeCmd(cfg *action.Configuration, logger log.Logger) *cobra.Command 
 
 			}
 			if client.ReleaseName == "" {
-				client.ReleaseName, _ = client.Name(ch)
+				client.ReleaseName, err = action.GetName(ch, "")
+				if err != nil {
+					return err
+				}
 				logger.Debugf("client.ReleaseName was empty, setting release name to %s", client.ReleaseName)
 			}
 
--- a/cmd/hypper/upgrade.go
+++ b/cmd/hypper/upgrade.go
@@ -123,13 +123,7 @@ func newUpgradeCmd(cfg *action.Configuration, logger log.Logger) *cobra.Command 
 					}
 					instClient := action.NewInstall(cfg)
 					// Set namespace for the install client
-					if settings.NamespaceFromFlag {
-						instClient.Namespace = settings.Namespace()
-					} else {
-						instClient.SetNamespace(ch, settings.Namespace())
-					}
-					// By this time client.Namespace has the correct namespace
-					instClient.Config.SetNamespace(instClient.Namespace)
+					action.SetNamespace(client, ch, settings.Namespace(), settings.NamespaceFromFlag)
 
 					instClient.CreateNamespace = createNamespace
 					instClient.ChartPathOptions = client.ChartPathOptions
@@ -167,16 +161,8 @@ func newUpgradeCmd(cfg *action.Configuration, logger log.Logger) *cobra.Command 
 			if err != nil {
 				return err
 			}
-
 			// Set namespace for the upgrade client
-			if settings.NamespaceFromFlag {
-				client.Namespace = settings.Namespace()
-			} else {
-				client.SetNamespace(ch, settings.Namespace())
-			}
-
-			// By this time client.Namespace has the correct namespace, so set it to the storage
-			client.Config.SetNamespace(client.Namespace)
+			action.SetNamespace(client, ch, settings.Namespace(), settings.NamespaceFromFlag)
 
 			rel, err := client.Run(client.ReleaseName, ch, vals)
 			if err != nil {
--- a/pkg/action/utils.go
+++ b/pkg/action/utils.go
@@ -1,0 +1,54 @@
+/*
+Copyright SUSE LLC.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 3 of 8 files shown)


Itxaka:
I KNOW I KNOW, LICENSE!!

pkg/action/utils.go
```diff
@@ -0,0 +1,42 @@
...
+	// setDefault is mainly used when we set the namespace via the cli flag --namespace and it has priority over
+	// everything else
+	if setDefault {
+		namespace = defaultns
+	}
+
+	switch i := x.(type) {
+	case *Install:
+		i.Namespace = namespace
+		i.Config.SetNamespace(namespace)
```
Itxaka, line 22: I could probably move this line to the last line after the case to avoid duplicating it, as we will run it always.

Or even remove the whole switch...case as my projection was to just call the `SetNamespace` on every action and then filter here instead, but it doesn't make much sense, does it?

Itxaka: well it may protect if we call `SetNamespace` from an action that has no `Namespace` as go wont catch that due to accepting an `interface` so it cant infer if it has a `Namespace` attribute....I think?????

Itxaka: I dont think we can call this from an action that has no `Namespace` so I think the whole switch...case needs to go.

Itxaka: ah seems that unless we do this, go will complain  of:

```
pkg/action/utils.go:44:3: x.Namespace undefined (type interface {} is interface with no methods)
pkg/action/utils.go:45:3: x.Config undefined (type interface {} is interface with no methods)

```

dragonchaser: This is some really confusing behaviour of go... I do not really like it, but I have no alternatives I could think of.

pkg/action/utils.go
```diff
@@ -0,0 +1,94 @@
...
+					namespace = val
+				}
+			}
+		}
+	}
+
+	switch i := x.(type) {
+	case *Install:
+		i.Namespace = namespace
+		i.Config.SetNamespace(namespace)
```
mattfarina, line 52: Here's an idea. This bit of functionality seems to repeat. Just this small part.

What about moving it to a `SetNamespace` property on `Upgrade` and `Install`. Then, create an interface with one method of `SetNamespace` that takes the namespace and does just this work. In the arguments for this function you can use that interface instead of `interface{}`. Then you don't have to do a type switch and can just call `i.SetNamespace(namespace)` here.

The compiler would make sure everything has the property.

You can also use composition to make that `SetNamespace` function once and embed it in the `Install` and `Upgrade` structs. Or, you could embed all of this that way.

Itxaka: Wouldn't that mean that:
 1 - We are still repeating code
 2 - We have 2 places to change the code if we find a bug

?

I like the second suggestion thought, seems more nice and easy to do with minimal changes!

Itxaka: Tried both suggestions, Im incapable of making it work as expected without messing it up or making it twice as complex...


Any suggestions or code examples? @mattfarina @viccuad @dragonchaser

pkg/action/utils.go
```diff
@@ -0,0 +1,54 @@
...
+	"helm.sh/helm/v3/pkg/chart"
+)
+
+// SetNamespace sets the Namespace that should be used based on annotations or fallback to default
+// both on the action and on the storage.
+// if setDefault is true then it will just set the default namespace
+// This will read the chart annotations. If no annotations, it leave the existing ns in the action.
+// targetNS can be either the default namesapce (usually "default") or the namespace passed via
+// cli flag
+func SetNamespace(x interface{}, chart *chart.Chart, targetNS string, NamespaceFromFlag bool) {
```
viccuad, line 31: `s/NamespaceFromFlag/setDefault/` (from the function comment)?
I would prefer `setDefault` or something similar instead of `namespaceFromFlag`, as API users may not use any "ns from flag", nor the API has any, so it may confuse them.

viccuad:
We need to refactor also https://github.com/rancher-sandbox/hypper/blob/510bdd2d29b2bfdebfb23b89100f128f372d514e/cmd/hypper/dependency_shared.go#L124-L130 now that it is in :D.

pkg/action/utils.go
```diff
@@ -52,3 +55,40 @@ func SetNamespace(x interface{}, chart *chart.Chart, targetNS string, NamespaceF
...
+	// cobra flags have been already stripped
+
+	flagsNotSet := func() error {
+		if nameTemplate != "" {
+			return errors.New("cannot set --name-template and also specify a name")
+		}
+		return nil
+	}
+
+	if len(args) > 2 {
```
viccuad, line 71: I'm not sure this is going to work out between `hypper install` and `hypper upgrade`. Install accepts 2 arguments, but upgrade only accepts 1, so `GetName()` can be reused.

pkg/action/utils.go
Itxaka: I used to have it as setDefault and changed it back to make it more clearer what it refer to... :rofl:

pkg/action/utils.go
Itxaka: it works because we are only using the args in the case if install. In that case we return on line 75, otherwise (upgrade or install with no name argument) we extract the name from annotations.

We could always end up removing the name param from install and just make it a flag like in upgrade and default to getting the name from the annotations always. Im not sure why upgrade was meant to be changed like that but not install..

viccuad: True! Thanks, now I see it :D.  In the case of upgrade, I suppose it is because in Helm you can't change the release name in upgrade, only upgrade. in install, we support setting a name.

I'm not completely sold on calling the function like `action.GetName(ch, "", make([]string, 0))`. It feels hackish, and people using the lib may not realize that's what it needs, without looking at the function code.

How about leaving install.Name() and upgrade.Name(), and creating an unexported function in action/utils.go to deduplicate the `if` in both?:
```
	if chart.Metadata.Annotations != nil {
		if val, ok := chart.Metadata.Annotations["hypper.cattle.io/release-name"]; ok {
			return val, nil
		}
		if val, ok := chart.Metadata.Annotations["catalog.cattle.io/release-name"]; ok {
			return val, nil
		}
	}
```

Or at the very least, documenting the new GetName() in a way that a lib user doesn't need to necessarily know about args or flags.

kkaempf: > I'm not completely sold on calling the function like action.GetName(ch, "", make([]string, 0)). It feels hackish, and people using the lib may not realize that's what it needs, without looking at the function code.

Can't we make this a variadic function so that `action.GetName(ch)` is sufficient in that case ?

Itxaka: done!

pkg/action/utils.go
Itxaka: reverted to `setDefault`

cmd/hypper/install.go
```diff
@@ -116,19 +116,12 @@ func runInstall(args []string, client *action.Install, valueOpts *values.Options
...
-
-	client.Config.SetNamespace(client.Namespace)
+	// Set namespace for the install client
+	action.SetNamespace(client, chartRequested, settings.Namespace(), settings.NamespaceFromFlag)
 
-	name, err := client.Name(chartRequested, args)
-	if err != nil {
-		return nil, err
+	if client.ReleaseName == "" {
+		client.ReleaseName, _ = action.GetName(chartRequested, client.NameTemplate, args...)
```
dragonchaser, line 123: Shouldn't we handle this error?

cmd/hypper/upgrade.go
```diff
@@ -106,7 +106,7 @@ func newUpgradeCmd(cfg *action.Configuration, logger log.Logger) *cobra.Command
 
 			}
 			if client.ReleaseName == "" {
-				client.ReleaseName, _ = client.Name(ch)
+				client.ReleaseName, _ = action.GetName(ch, "")
```
dragonchaser, line 109: error handling?

dragonchaser (changes requested):

cmd/hypper/install.go
Itxaka: good catch!

Itxaka: fixed!

cmd/hypper/upgrade.go
Itxaka: Fixed!

dragonchaser (approved):

mattfarina (approved):

Issue #95, opened 2021-03-16
Releases


mattfarina:
Setup releases of hypper for mac, linux, and windows on appropriate architectures using something like https://goreleaser.com/.

PR #96, opened 2021-03-17
use goreleaser to provide hypper builds

Itxaka:
Uses the goreleaser action to build and
provide hypper builds for macos, linux, windows
in an automated fashion

Fixes: #95 

### DIFF
--- a/.github/workflows/releaser.yml
+++ b/.github/workflows/releaser.yml
@@ -1,0 +1,26 @@
+name: goreleaser
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+jobs:
+  goreleaser:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.16
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v2
+        with:
+          version: latest
+          args: release --rm-dist
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,0 +1,39 @@
+# Make sure to check the documentation at http://goreleaser.com
+builds:
+  - main: ./cmd/hypper
+    ldflags:
+      - -w -s
+      - -X github.com/rancher-sandbox/hypper/internal/version.version={{.Tag}}
+      - -X github.com/rancher-sandbox/hypper/internal/version.gitCommit={{.Commit}}
+    env:
+      - CGO_ENABLED=0
+    goos:
+      - linux
+      - windows
+      - darwin
+    goarch:
+      - amd64
+      - arm
+      - arm64
+source:
+  enabled: true
+  name_template: '{{ .ProjectName }}-{{ .Tag }}-source'
+archives:
+    # Default template uses underscores instead of -
+  - name_template: "{{ .ProjectName }}-{{ .Tag }}-{{ .Os }}-{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}"
+    replacements:
+      darwin: Darwin
+      linux: Linux
+      windows: Windows
+      386: i386
+      amd64: x86_64
+checksum:
+  name_template: '{{ .ProjectName }}-{{ .Tag }}-checksums.txt'
+snapshot:
+  name_template: "{{ .Tag }}-next"
+changelog:
+  sort: asc
+  filters:
+    exclude:
+      - '^docs:'
+      - '^test:'


viccuad (commented):
Wow! Didn't know about goreleaser, pretty cool! And nice set up with GHA!

:thinking: Could we depend on it for local and CI builds? I would prefer to have the same stack locally, on CI, and on CD; if not it feels that we are not doing CI and CD correctly (It scares me that we can test something locally, submit PRs that pass tests and CIs, and then the releaesd build can behave completely different because we forgot config or the build process is different).

Itxaka:
@viccuad it runs on local with no issues (you only need to run --snapshot --skip-publish so it doesnt try to publish the artifacts)

In any case, other than for testing, Im not sure there is a necessity to run it locally as its only executed when new tags are pushed. Plus it needs a GITHUB_TOKEN in order to publish, so IMO its easier to skip any local facilities in the Makefile.

Itxaka:
Plus it would be good for the whole process of releasing to have one source of truth and not be able to mix or overwrite releases because we are local testing :D

viccuad (dismissed):
Thought about using goreleaser in `make build` target, to have a consistent stack. Tried it locally with `goreleaser build --skip-validate --rm-dist` and of course it needs a full minute (and several threads) to build all the releases for all OSes. Whiich is too much time for development. We could load different configs when working locally, and so on, but it gets complicated.

Approving then :).

Itxaka:
Sorry, I updated it to make the checksums and archive names more in sync with other projects in the same space.

Now uses `-` instead of `_` in teh archive names and adds the `v` in front of the version for the archives. Also updates the checksum name file as well.

viccuad (approved):

Itxaka:
waiting for some tests offline before merging...

Itxaka:
ok sorry about that, now it produces the same name for all archives, including checksums and source

viccuad (dismissed):

viccuad:
I did some stalking and saw https://github.com/Itxaka/hypper/releases :D. Super nice!

.github/workflows/releaser.yml
```diff
@@ -0,0 +1,26 @@
+name: goreleaser
+
+on:
+  push:
+    tags:
+      - '*'
```
dragonchaser, line 6: If I read that correctly this will create a release on every tag, wouldn't it be better to only release on `v.*`?

.goreleaser.yml
```diff
@@ -0,0 +1,35 @@
...
+      386: i386
+      amd64: x86_64
+checksum:
+  name_template: '{{ .ProjectName }}-v{{ .Version }}-checksums.txt'
+snapshot:
+  name_template: "{{ .Tag }}-next"
+changelog:
+  sort: asc
+  filters:
+    exclude:
```
dragonchaser, line 33: we might also want to exclude the third party folders...

dragonchaser (changes requested):

.github/workflows/releaser.yml
Itxaka: That is a good point

.goreleaser.yml
Itxaka: You mean the `internal/third_party` dir? That is still part of hypper isnt it? And if there is a bug that is fixed by changing something in the folder (updating a file, adding a new one) dont we want to have it reflected on the changelog?

dragonchaser: IMHO I would say no, `third_party` should only contain raw third party code without changes, everything else would be moved to `pkg` and we would reflect that in the changelog.

Itxaka: sure, but for example we currently have a `urlutil.go` file there. If there is a bug in that file (upstream at helm) and we update it locally to match the updated file...should that be in the changelog in order to identify in which version was that change introduced?

mattfarina: The only changes to the third party folder now would be things like us syncing changes from upstream or making an intentional change, right? Shouldn't those be in the changelog?

.github/workflows/releaser.yml
mattfarina: Will we have any tags that are not v[semver]? We need that format for Go and our internal library. The '*' may work for now.

.goreleaser.yml
Itxaka: I think so, because in the end its part of what we have in our repo, so its our responsability. I would not exclude those from the changelog.

.github/workflows/releaser.yml
Itxaka: I guess that its a safe option in case we mess up with the tags? I dont mind adding that.

dragonchaser: Well we could have hotfixes that are tagged and do not have an automated release...

.goreleaser.yml
dragonchaser: @Itxaka I guess you are right, just include them...

.github/workflows/releaser.yml
Itxaka: updated to add the `v.*` filter to be safe!

viccuad (dismissed):

Itxaka:
added missing `ldflags`

@viccuad Those are the ones being set by goreleaser and work the same as the ones in the Makefile :D

viccuad (approved):
Nice! :rocket:

Itxaka:
lets gooooo

PR #98, opened 2021-03-18
Add version command

Itxaka:
To know the version of hypper and commit


Itxaka:
I think I migth have found a bug in the log library as per the failing test due to strange output when using a template for version print.

Itxaka:
@mattfarina So if I build this PR and use version it goes liek this:

```
$ bin/hypper version --template='{{.Version}}'              
v0.0.1
```

That is correct, but if I add something in front of the Version the logger prints it in 2 lines for some reason:

```
$ bin/hypper version --template='Version {{.Version}}'
Version 
v0.0.1
```

Uing a simple `io.Writer` instead makes it print the proper output in one line so...either Im using it wrong or there is some kind of bug in the log lib??

cmd/hypper/version.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+func (o *versionOptions) run(wr io.Writer) error {
+	if o.template != "" {
+		tt, err := template.New("_").Parse(o.template)
+		if err != nil {
+			return err
+		}
+		return tt.Execute(wr, version.Get())
+	}
```
mattfarina, line 87: I figured out why we are seeing cases of a template string getting newlines in it.

The log based writer adds a new line each time `.Write()` is called. It's meant to write to logs and adds a new line if one is not present.

When the template executor is creating the output it would take a template like `"Foo {{  .Version }} Bar"` and write it in 3 parts. First is would send `Foo ` to the writer. Then it would calculate the string for `{{  .Version }}` and send it to the writer. Then it would sent ` Bar` to the writer.

Each time it does that the log writer adds a new line. It is expecting to write a single log line each time.

There is a rather simple work around for this. Write the template output to a buffer and send that to the writer. Like the following...

```go
buf := &bytes.Buffer{}
tt.Execute(buf, version.Get())
io.Copy(wr, buf)
```

viccuad (approved):

internal/version/version.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+var (
+	// version is the current version of Hypper.
+	// Update this whenever making a new release.
+	// The version is of the format Major.Minor.Patch[-Prerelease][+BuildMetadata]
+	//
+	// Increment major number for new feature additions and behavioral changes.
+	// Increment minor number for bug fixes and performance enhancements.
+	version = "v0.0.1"
```
viccuad, line 31: silly question, does goreleaser use this?

Itxaka: What goreleaser does is sets the override for this using the ldflags like we do here on the Makefile[0]

Of course, I cant add that to the goreleaser until the flags are set here...I think?? So this is actually blocking the goreleaser PR and I just noticed it!


[0] https://github.com/rancher-sandbox/hypper/pull/98/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R22

Itxaka: its easier to see it run:

```

# itxaka @ zeus in ~/projects/hypper on git:add_version o [15:39:35] C:128
$ git tag v0.0.12

# itxaka @ zeus in ~/projects/hypper on git:add_version o [15:39:39] 
$ make build
go fmt ./...
internal/version/version.go
go vet ./...
go build  -tags '' -ldflags '-w -s -X github.com/rancher-sandbox/hypper/internal/version.version=v0.0.12 -X github.com/rancher-sandbox/hypper/internal/version.gitCommit=0f51f331425a55878bd69872d8a1ee42bef5441e' -o '/home/itxaka/projects/hypper/bin'/"hypper" ./cmd/hypper

# itxaka @ zeus in ~/projects/hypper on git:add_version x [15:39:50] 
$ bin/hypper version                                  
version.BuildInfo{Version:"v0.0.12", GitCommit:"0f51f331425a55878bd69872d8a1ee42bef5441e", GoVersion:"go1.16"}

```

In this case the makefile is setting `X github.com/rancher-sandbox/hypper/internal/version.version=v0.0.12` so the end binary has that version. Goreleaser will do the same by setting the same flag on build

viccuad: many thanks for the extended explanation and example!

Issue #57, opened 2021-03-01
hypper lint


mattfarina:
In addition to Helm linting rules it should:

- look at the metadata in the annotations and make sure it is properly formatted
- Check dependent charts to make sure they are properly setup for their use. For example, if a shared dependency is specified the release name and namespace should be set for that chart.

PR #109, opened 2021-04-05
Add lint command

Itxaka:
Adds the lint command to hypper and rules for our
annotations.

This is run *after* the normal helm rules as to
cover as much terrain as possible so on top of
the usual helm rules we can add our own rules like
in this patch.

Currently the hypper rules are on Info level and cover
both release-name and namespace annotations

partially-fixes: #57 


viccuad:
I haven't reviewed in depth, but could we also check for `annotations. hypper.cattle.io/shared-dependencies` too? (as listed in https://github.com/rancher-sandbox/hypper/issues/30. The current implementation unmarshals here https://github.com/rancher-sandbox/hypper/blob/main/pkg/action/dependency_shared.go#L94).

Itxaka:
Added user docs!

docs/user/howto/lintchart.md
```diff
@@ -0,0 +1,61 @@
+## How to lint a hypper chart
+
+We just created or modified our chart or add annotations to take advantage of Hypper features but...how do we know those annotations are correct?
+
+
+Its very simple, just run `hypper lint` against your chart, and it will verify that the chart is well-formed.
+
+`hypper lint` run both Helm checks and Hypper checks against the chart.
```
viccuad, line 8:
```suggestion
`hypper lint` runs both Helm checks and Hypper checks against the chart.
```

pkg/action/lint.go
```diff
@@ -0,0 +1,112 @@
+/*
+Copyright The Helm Authors, Suse LLC
```
viccuad, line 2: Nit: maybe SUSE should be in uppercase (we should fix the linter, because right now it's passing as it is).

pkg/lint/lint.go
```diff
@@ -0,0 +1,33 @@
+/*
+Copyright Suse LLC
```
viccuad, line 2: nit: SUSE should be in uppercase.

pkg/lint/lint.go
```diff
@@ -0,0 +1,33 @@
...
+	"path/filepath"
+)
+
+// All runs all of the available linters on the given base directory.
+func All(basedir string) support.Linter {
+	// Using abs path to get directory context
+	chartDir, _ := filepath.Abs(basedir)
+
+	linter := support.Linter{ChartDir: chartDir}
+	rules.Annotations(&linter)
```
viccuad, line 31: Shouldn't we be extending Helm's `pkg/lint`, instead of having our own without the Helm checks?
That way it would be more upstreamable.

I would go with having a `pkg/lint/lint.go.All()` that calls the Helm rules, by importing them. And then `pkg/lint/rules/ ` would keep containing only Hypper rules.
(This means that pkg/action/lint.go doesn't need to call Helm's linters and then Hypper's, of course).

cmd/hypper/lint.go
```diff
@@ -0,0 +1,128 @@
...
+							} else if strings.HasSuffix(path, ".tgz") || strings.HasSuffix(path, ".tar.gz") {
+								paths = append(paths, path)
+							}
+						}
+						return nil
+					})
+				}
+			}
+
+			client.Namespace = settings.Namespace()
```
viccuad, line 73: open question, is this enough? do we need to change the `client.Config` too?

pkg/action/lint.go
viccuad: (also, other headers too).

viccuad (commented):
Looks nice! wooho! I love the linters :).

pkg/lint/lint.go
Itxaka: Well the `pkg/lint` from helm only has the `All` method, everything is under the `pkg/action/lint` and Im extending that on `pkg/action/lint` already.

Plus changing the `All` method means modifying the signature of that so a full reqork of other pieces.

Not sure how that would be better to what is currently in the PR. Maybe Im just not understanding what you mean?

cmd/hypper/lint.go
Itxaka: for linting yes, as it doesn't really do anything with the client namespace, everything is local. Probably would be able to even remove it...but its left like that because its the same upstream

viccuad: true! thanks for the explanation :)

pkg/lint/lint.go
viccuad: My idea was that `pkg/lint.All()` should call all the Helm rules + the Hypper ones, so:
https://github.com/helm/helm/blob/167aac70832d3a384f65f9745335e9fb40169dc2/pkg/lint/lint.go#L32-L35
+ `rules.Annotations(&linter)`.

That way we don't need to iterate twice for Helm and Hypper rules in `pkg/action/lint.go.Run()`. In fact, we may not need to re-implement `pkg/action/lint.go.Run()` and `pkg/action/lint.go.lintAll()` at all, and then we would be dropping its tests too.

That way we only maintain the Hypper specific tests for `pkg/lint/rules`. And it may very well be that we only need changes there in the future even. Which would make upstreaming `pkg/lint/rules/*` easier.

At the very least, I think we should just call all the Helm rules in `pkg/lint.All()` instead of `pkg/action/lint.go.Run()`.

Itxaka: but `pkg/action/lint.go.lintChart` is the one calling the `pkg/lint/lint.go.All()` ?

So we need to override the `pkg/action/lint::Linter.Run` in any case so we can run our rules. Otherwise there is no of hijacking the linter


I still dont see how to make it otherwise. 
The only way is removing the call to the helm linter (which I would not do, reasons below), importing the rules, and calling them in our side just above the hypper ones (pkg/lint/lint.go.All())

My reasons for not wanting to bring the rules are that if something changes in the linter internal code or rules then we may miss it, while the API for the linter is not gonna change ever, we call the linter, it return the results, we dont care what they do internally, if they added, removed or dropped rules, if they now do some mangling to their rules, we only care of running our rules.

Im not against it, its just doesn't seem like a good idea.

Itxaka: updated the PR so it now does this.

pkg/lint/rules/annotations.go
```diff
@@ -0,0 +1,81 @@
...
+		return errors.New("Setting hypper.cattle.io/shared-dependencies in annotations is recommended")
+	}
+	return nil
+}
+
+// validateChartHypperSharedDepsCorrect checks that shared deps are in the correct format
+func validateChartHypperSharedDepsCorrect(chart *helmChart.Metadata) error {
+	depList := chart.Annotations["hypper.cattle.io/shared-dependencies"]
+	var yamlDeps []*helmChart.Dependency
+	if err := yaml.Unmarshal([]byte(depList), &yamlDeps); err != nil {
```
viccuad, line 77: We should use UnmarshallStrict instead, the yaml must match Dependency.

pkg/lint/rules/testdata/goodchart/Chart.yaml
```diff
@@ -0,0 +1,12 @@
...
+description: A Helm chart for Kubernetes
+type: application
+version: 0.1.0
+appVersion: "1.16.0"
+icon: "http://example.com/icon.png"
+annotations:
+  "hypper.cattle.io/release-name": "release"
+  "hypper.cattle.io/namespace": "namespace"
+  "hypper.cattle.io/shared-dependencies": |
+    - key: value
```
viccuad, line 12: This shouldn't pass the test, it should be `- name: value`.

pkg/lint/rules/annotations.go
viccuad: As we do here: https://github.com/rancher-sandbox/hypper/blob/441961614688aa0f338c2a22236d56cf6038f729/pkg/action/dependency_shared.go#L94

Which takes the struct annotations in Helm's `pkg/chart/dependency.go.Dependency`.

pkg/lint/rules/testdata/goodchart/Chart.yaml
Itxaka: mmh, isnt this just checking if the `hypper.cattle.io/shared-dependencies` contains a valid yaml? 

The test should probably be expanded to not only validate that yaml but also for each entry in there try to load a `Dependency` object with those values.

I kind of wanted to make this simple, expanding it over time as its not an urgent matter and Im not sure if the shared stuff is final (like it could change format?? I dont know sorry)

Itxaka: oh wait, is this because I messed up the Unmarshall in line 77??? WOW!Good catch, I almost messed up!

cmd/hypper/lint_test.go
```diff
@@ -0,0 +1,40 @@
+// +build linux
```
viccuad, line 1: mmh, we should run the test on all OSes :/. TBF I don't see why the filepath walking would fail in Win, maybe we will be lucky.

docs/user/howto/lintchart.md
```diff
@@ -0,0 +1,61 @@
+## How to lint a hypper chart
+
+We just created or modified our chart or add annotations to take advantage of Hypper features but...how do we know those annotations are correct?
```
viccuad, line 3:
```suggestion
We just created or modified our chart or added annotations to take advantage of Hypper features but...how do we know those annotations are correct?
```

docs/user/howto/lintchart.md
```diff
@@ -0,0 +1,61 @@
...
+
+
+Its very simple, just run `hypper lint` against your chart, and it will verify that the chart is well-formed.
```
viccuad, line 6:
```suggestion
It's very simple, just run `hypper lint` against your chart, and it will verify that the chart is well-formed.
```

cmd/hypper/lint_test.go
Itxaka: I think the output of the not found error is slightly different, so looking for the error text did not work, nor I have a easy available windows machine to test further :)

pkg/action/lint_test.go
```diff
@@ -0,0 +1,152 @@
+// +build linux
```
viccuad, line 1: I wonder if we can just drop this too.

pkg/lint/lint.go
viccuad: yay! thanks!

cmd/hypper/lint_test.go
viccuad: Maybe it's possible to do what we did here:

https://github.com/rancher-sandbox/hypper/blob/441961614688aa0f338c2a22236d56cf6038f729/cmd/hypper/dependency_shared_test.go#L66-L69

I ended up trying a couple commits until I got the output correctly :P.

Itxaka: oh god, the linting error is because of the windows paths vs linux paths....


dont do this to me.

pkg/action/lint_test.go
Itxaka: done!

cmd/hypper/lint_test.go
Itxaka: done!

viccuad (approved):
WoOooohoo!

Issue #78, opened 2021-03-09
`hypper search`


dragonchaser:
Implement the `hypper search` command to be able to search through repositories similar to `helm search`.

PR #113, opened 2021-04-07
Add search command

Itxaka:
Adds the helm repo search command to hypper as just hypper search

Under helm you can search on the repos or on their own hub. This
patch only implements the repo search as seems the most useful
and bringing the hub requires a lot of code

This patch also extracts the search code into a new pkg as under
helm the search code is mostly crammed under cmd which means you cannot
use it anywhere, meanwhile this code can be reused by third party
clients to implement their own search frontend or automate it.

Fixes: #78 

### DIFF
--- a/cmd/hypper/search.go
+++ b/cmd/hypper/search.go
@@ -1,0 +1,92 @@
+/*
+Copyright The Helm Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"github.com/Masterminds/log-go"
+	"github.com/rancher-sandbox/hypper/pkg/search"
+	"github.com/spf13/cobra"
+)
+
+const searchDesc = `
+Search provides the ability to search for Hypper charts in the various places
+they can be stored including repositories you have added.
+Use search subcommands to search different locations for charts.
+`
+
+const searchRepoDesc = `
+Search reads through all of the repositories configured on the system, and
+looks for matches. Search of these repositories uses the metadata stored on
+the system.
+
+It will display the latest stable versions of the charts found. If you
+specify the --devel flag, the output will include pre-release versions.
+If you want to search using a version constraint, use --version.
+
+Examples:
+
+    # Search for stable release versions matching the keyword "nginx"
+    $ hypper search repo nginx
+
+    # Search for release versions matching the keyword "nginx", including pre-release versions
+    $ hypper search repo nginx --devel
+
+    # Search for the latest stable release for nginx-ingress with a major version of 1
+    $ hypper search repo nginx-ingress --version ^1.0.0
+
+Repositories are managed with 'hypper repo' commands.
+`
+
+// newSearchCmd is the generic search command that can contain subcommands
+func newSearchCmd(logger log.Logger) *cobra.Command {
+
+	cmd := &cobra.Command{
+		Use:   "search [keyword]",
+		Short: "search for a keyword in charts",
+		Long:  searchDesc,
+	}
+
+	cmd.AddCommand(newSearchRepoCmd(logger))
+
+	return cmd
+}
+
+// newSearchRepoCmd search on repos
+func newSearchRepoCmd(logger log.Logger) *cobra.Command {
+	o := &search.RepoOptions{}
+
+	cmd := &cobra.Command{
+		Use:   "repo [keyword]",
+		Short: "search repositories for a keyword in charts",
+		Long:  searchRepoDesc,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			o.RepoFile = settings.RepositoryConfig
+			o.RepoCacheDir = settings.RepositoryCache
+			return o.Run(logger, args)
+		},
+	}
+
+	f := cmd.Flags()
+	f.BoolVarP(&o.Regexp, "regexp", "r", false, "use regular expressions for searching repositories you have added")
+	f.BoolVarP(&o.Versions, "versions", "l", false, "show the long listing, with each version of each chart on its own line, for repositories you have added")
+	f.BoolVar(&o.Devel, "devel", false, "use development versions (alpha, beta, and release candidate releases), too. Equivalent to version '>0.0.0-0'. If --version is set, this is ignored")
+	f.StringVar(&o.Version, "version", "", "search using semantic versioning constraints on repositories you have added")
+	f.UintVar(&o.MaxColWidth, "max-col-width", 50, "maximum column width for output table")
+	bindOutputFlag(cmd, &o.OutputFormat)
+
+	return cmd
+}
--- a/docs/user/howto/search.md
+++ b/docs/user/howto/search.md
@@ -1,0 +1,50 @@
+## Searching for charts in repos
+
+Search reads through all the repositories configured on the system, and looks for matches. Searching on these repositories uses the metadata stored on the system.
+
+It will display the latest stable versions of the charts found.
+
+```shell
+$ hypper search repo fleet
+NAME              	CHART VERSION	APP VERSION	DESCRIPTION                            
+hypper/fleet      	0.3.500      	0.3.5      	Fleet Manager - GitOps at Scale        
+hypper/fleet-agent	0.3.500      	0.3.5      	Fleet Manager Agent - GitOps at Scale  
+hypper/fleet-crd  	0.3.500      	0.3.5      	Fleet Manager CustomResourceDefinitions
+```
+
+By default, it will only show the latest version that match the keyword used. In order to see all versions you can use the `-l` flag.
+
+```shell
+$ hypper search repo fleet-agent -l
+NAME              	CHART VERSION	APP VERSION	DESCRIPTION                          
+hypper/fleet-agent	0.3.500      	0.3.5      	Fleet Manager Agent - GitOps at Scale
+hypper/fleet-agent	0.1.500      	0.1.5      	Fleet Manager Agent - GitOps at Scale
+```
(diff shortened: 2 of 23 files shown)


cmd/hypper/search.go
```diff
@@ -0,0 +1,71 @@
...
+    $ hypper search nginx-ingress --version ^1.0.0
+
+Repositories are managed with 'hypper repo' commands.
+`
+
+func newSearchCmd(logger log.Logger) *cobra.Command {
+	o := &search.RepoOptions{}
+
+	cmd := &cobra.Command{
+		Use:   "search [keyword]",
```
viccuad, line 52: I guess this is about preferences, but here we are implementing the new command as `hypper search [foo]`, instead of `hypper search repo [foo]`, which would be analogous to helm's `helm search repo [foo]`. Is that what we want? (just in case, as people may expect it on the latter).

pkg/search/search.go
```diff
@@ -0,0 +1,214 @@
...
+	AppVersion  string `json:"app_version"`
+	Description string `json:"description"`
+}
+
+type repoSearchWriter struct {
+	results     []*search.Result
+	columnWidth uint
+}
+
+func (r *repoSearchWriter) WriteTable(out io.Writer) error {
```
viccuad, line 171: Is it function used somewhere? (also, I don't see tests for it)

viccuad (commented):
Nice! having the search functionality in the libs sounds awesome :)

I have only done a fast review so far. Could we first double-check on the exported functions and methods that will be in the lib API? For example, maybe we don't need to expose WriteJSON() and WriteYAML(). Also, could we add comments to the exported ones? (it would help me with the review).

pkg/search/search.go
Itxaka: This is part of  the `helm/pkg/cli/output/output.go` and called by line 91 in `search/search.go` based on the `RepoOptions.OutputFormat` which is set via command line flag as part of the `bindOutputFlag` :)

A bit convoluted, I agree but this is basically copied from helm.

Itxaka:
> For example, maybe we don't need to expose WriteJSON() and WriteYAML(). 

Unfortunately due to the convoluted way in which those functions are called they need to be exposed because `repoSearchWriter` needs to implement the `Writer` interface in order to be called.

cmd/hypper/search.go
Itxaka: I could just modify it to be `repo search` in case we want to expand it further because I agree, coming from helm you would expect it to work like that

viccuad: I would be happy with the command being `hypper search repo`, just to mimic Helm's ui. No need to bring the `hypper search hub`, indeed. Who knows, maybe in the future we will have `hypper search cache` :).

pkg/search/search.go
viccuad: Thanks for the explanation!

pkg/search/search.go
```diff
@@ -0,0 +1,226 @@
...
+		return nil, errors.New("no repositories configured")
+	}
+
+	i := search.NewIndex()
+	for _, re := range rf.Repositories {
+		n := re.Name
+		f := filepath.Join(o.RepoCacheDir, helmpath.CacheIndexFile(n))
+		ind, err := helmRepo.LoadIndexFile(f) // Needs to be helm type as we call the addrepo from helm directly
+		if err != nil {
+			log.Warn("Repo %q is corrupt or missing. Try 'helm repo update'.", n)
```
viccuad, line 155:
```suggestion
			log.Warn("Repo %q is corrupt or missing. Try 'hypper repo update'.", n)
```

Itxaka:
user docs added!

pkg/search/search.go
```diff
@@ -0,0 +1,226 @@
...
+
+// buildIndex loads the repos to add them to the index search
+func (o *RepoOptions) buildIndex() (*search.Index, error) {
+	// Load the repositories.yaml
+	rf, err := repo.LoadFile(o.RepoFile)
+	if isNotExist(err) || len(rf.Repositories) == 0 {
+		return nil, errors.New("no repositories configured")
+	}
+
+	i := search.NewIndex()
```
viccuad, line 149: `search.NewIndex()` comes from `cmd/helm/search`. I'm a bit un-easy with that. I wonder, should we port it to `hypper/pkg/`, too?

pkg/search/search.go
```diff
@@ -0,0 +1,226 @@
...
+		n := re.Name
+		f := filepath.Join(o.RepoCacheDir, helmpath.CacheIndexFile(n))
+		ind, err := helmRepo.LoadIndexFile(f) // Needs to be helm type as we call the addrepo from helm directly
+		if err != nil {
+			log.Warn("Repo %q is corrupt or missing. Try 'hypper repo update'.", n)
+			log.Warn("%s", err)
+			continue
+		}
+
+		i.AddRepo(n, ind, o.Versions || len(o.Version) > 0)
```
viccuad, line 160: same with `i.AddRepo()`, comes from `cmd/helm/search`.

pkg/search/search.go
viccuad: I remember reading a doc file from Helm, listing the things covered under backwards compatibility/semver. I think `pkg/` API is covered, as well as  cmd cli UI is covered, but I don't think `cmd/` API is covered though. But I can't find the doc right now, grr.

cmd/hypper/search.go
viccuad: Now this seems to be addressed, by looking at the tests and newSearchRepoCmd(), but I think we missed the `const searchRepoDesc` examples.

pkg/search/search.go
Itxaka: that is a good point, I didnt noticed that.

Im not even sure why they are cramming everything in the cmd part...

cmd/hypper/search.go
Itxaka: not as far as I can see, the same text is in place.

viccuad: shouldn't they be `$ hypper search repo nginx` instead of `$ hypper repo search nginx`?

pkg/search/search.go
Itxaka: ah found it:

```
Go libraries inside of pkg/ SHOULD remain backward compatible, though code inside of cmd/ and internal/ may be changed from release to release without notice.```

Comment (Itxaka) [reply pkg/search/search.go]: done!

File: pkg/search/search.go (hunk 5 above)
Comment (Itxaka) [reply pkg/search/search.go:158]: done!

File: cmd/hypper/search.go (hunk 1 above)
Comment (Itxaka) [reply cmd/hypper/search.go:58]: woops, dyslexia galore! Fixed!

File: cmd/hypper/search.go [hunk 6]
```diff
@@ -0,0 +1,92 @@
...
+package main
+
+import (
+	"github.com/Masterminds/log-go"
+	"github.com/rancher-sandbox/hypper/pkg/search"
+	"github.com/spf13/cobra"
+)
+
+const searchDesc = `
+Search provides the ability to search for Helm charts in the various places
```
viccuad, line 26:
```suggestion
Search provides the ability to search for Hypper charts in the various places
```
Just realized..

Itxaka: woops, if you accept the suggestion it auto resolves, sorry about that!

viccuad: yep, the missing `Signed-off-by: Víctor Cuadrado Juan <vcuadradojuan@suse.de>` is a pity :/. Can you amend the last commit?

viccuad (dismissed):
Approving, as as we talked on the chat, is gonna get rebased.

cmd/hypper/search.go
viccuad: thanks! :)

docs/user/howto/search.md
```diff
@@ -0,0 +1,50 @@
+## Searching for charts in repos
+
+Search reads through all the repositories configured on the system, and looks for matches. Search of these repositories uses the metadata stored on the system.
```
viccuad, line 3: Nit: (maybe? I just had trouble with the phrase)

```suggestion
Search reads through all the repositories configured on the system, and looks for matches. Searching on these repositories uses the metadata stored on the system.
```

docs/user/howto/search.md
```diff
@@ -0,0 +1,50 @@
...
+It's also possible to pass the `--regexp` flag to use regexp in the search.
+
+```shell
+$ hypper search repo "fleet-" --regexp
+NAME              	CHART VERSION	APP VERSION	DESCRIPTION                            
+hypper/fleet-agent	0.3.500      	0.3.5      	Fleet Manager Agent - GitOps at Scale  
+hypper/fleet-crd  	0.3.500      	0.3.5      	Fleet Manager CustomResourceDefinitions
+```
+
+If you want to search for development charts (only stable charts are show by default), use the `--devel` flag to show those development charts.
```
viccuad, line 43:
```suggestion
If you want to search for development charts (only stable charts are shown by default), use the `--devel` flag to show those development charts.
```

pkg/search/search.go
```diff
@@ -0,0 +1,224 @@
...
+	Regexp       bool
+	Devel        bool
+	Version      string
+	MaxColWidth  uint
+	RepoFile     string
+	RepoCacheDir string
+	OutputFormat output.Format
+}
+
+// Run search and prints the found charts based of the filters
```
viccuad, line 59:
```suggestion
// Run searchs and prints the found charts based of the filters
```

viccuad (dismissed):
Wooohoo! nice, and with the `cmd/helm/search` things in our own `pkg/search`!

Approving, feel free the consider my reviews as passed if/when typos get pushed!

docs/user/howto/search.md
Itxaka: shown is past participle. Is that what we want?

viccuad: You made me google :sweat_smile: , it just sounded bad here. We are doing passive voice ("the charts are shown"), which is done with past participle.

Itxaka: ok, so then its correct in past snse, good, let me change it!

viccuad (approved):

Itxaka:
rebasing for the conflicts...

Itxaka:
was already approved by victor, I had to rebase due to merge conflicts on the root.go file (because lint added a command there and we are also adding a command)

So merging :)

Issue #93, opened 2021-03-16
Filter based on annotations


mattfarina:
We need a method to filter the available charts based on annotations. This is useful to rancher so that rancher can have one repository and different versions of charts that work with different versions of rancher.

There are a few ways we could go about this. When this ticket is taken there can be further discussion.

1. An annotation whose value is a semver. Note, this would not be tied to the hypper version but a version that could be passed into hypper
2. The go API providing a method to filter repositories based on metadata in charts
3. Some combination of the two that provides a means to do advanced filtering via the cli (semver aware) but not specific to rancher

We want to make this feature something in the API (like a filtering callback function) and possibly something exposed to the CLI that rancher can use but it's not rancher specific.

Itxaka:
blocked by #113

mattfarina:
@Itxaka how does search block this? Can search accept the same information and use that as part of its filtering? I realize this would deviate from upstream Helm.

Itxaka:
> @Itxaka how does search block this? Can search accept the same information and use that as part of its filtering? I realize this would deviate from upstream Helm.

Well, this build on top of the search command and as long as we dont have the search implemented on hypper we cannot implement this.

So more like `adding the search command to hypper` is blocking this :D

Itxaka:
@mattfarina I just noticed that I didnt had you on the PR that tackled this, so probably you only saw my comment in this issue, sorry about that :D

mattfarina:
#114 Implements part of it.

Still needed is the ability to handle version ranges and an API based method

viccuad:
@mattfarina I'm having trouble parsing point 1 from the card. Is it correct, that the intent is to have an API method to filter charts based on a semver parameter? And preferably do that with an API method that accepts filters as callback functions.

PR #114, opened 2021-04-07
Add filtering by annotations on search

Itxaka:
Allows to add filtering by annotations by passing the -a flag and the
annotation in the format "KEY=VALUE"
    
So you can search like this:

`hypper search repo -a "catalog.cattle.io/certified=rancher"`

And it will only show charts that have that exact annotation.

Requires #113 

### DIFF
--- a/cmd/hypper/search_test.go
+++ b/cmd/hypper/search_test.go
@@ -72,7 +72,32 @@ func TestSearchRepositoriesCmd(t *testing.T) {
 		name:   "search for 'alpine', expect valid yaml output",
 		cmd:    "search repo alpine --output yaml",
 		golden: "output/search-output-yaml.txt",
-	}}
+	}, {
+		name:   "search for hypper, expect latest version 0.2.0",
+		cmd:    "search repo hypper --output yaml",
+		golden: "output/search-hypper-latest.txt",
+	}, { // I use yaml for the output because its easier to create the golden files, sue me :p
+		// Should find version 0.1.0 as that is the only one with that annotation
+		name:   "search for hypper with annotation hypper.cattle.io/namespace:hyppernamespace",
+		cmd:    "search repo hypper -a 'hypper.cattle.io/namespace=hyppernamespace' --output yaml",
+		golden: "output/search-hypper-annotation.txt",
+	}, {
+		// Should find latest version 0.2.0 with that annotation
+		name:   "search for hypper with regexp annotation hypper.cattle.io/release-name:hypper*",
+		cmd:    "search repo hypper --regexp -a 'hypper.cattle.io/release-name=hypper*' --output yaml",
+		golden: "output/search-hypper-annotation-regexp.txt",
+	}, {
+		// Should find that exact version with the annotation, not the latest
+		name:   "search for hypper with version and annotation hypper.cattle.io/release-name:hyppername",
+		cmd:    "search repo hypper --version 0.1.0 -a 'hypper.cattle.io/release-name=hyppername' --output yaml",
+		golden: "output/search-hypper-annotation-version.txt",
+	}, {
+		// Now everything at the same time! regexp! version! annotations!
+		name:   "search for hypper with version and annotation hypper.cattle.io/release-name:hyppername",
+		cmd:    "search repo hypper --regexp --version 0.1.0 -a 'hypper.cattle.io/release-name=^hypper' --output yaml",
+		golden: "output/search-hypper-annotation-version-regexp.txt",
+	},
+	}
 
 	settings.Debug = true
 	defer func() { settings.Debug = false }()
--- a/docs/user/howto/searchannotations.md
+++ b/docs/user/howto/searchannotations.md
@@ -1,0 +1,34 @@
+## Searching for charts in repos using annotations
+
+Apart from having the normal [search](docs/user/howto/search.md) in hypper, we support searching for specific annotations in a search.
+
+Using the `-a` flag in `hypper search repo` allows to filter the returned charts based on the KEY=VALUE values in annotations.
+
+```shell
+$ hypper search repo -a 'hypper.cattle.io/release-name=fleet'       
+NAME        	CHART VERSION	APP VERSION	DESCRIPTION                    
+hypper/fleet	0.3.500      	0.3.5      	Fleet Manager - GitOps at Scale
+
+```
+
+
+We also support the `--regexp` flag when using annotations, so you can search for the VALUE with a regexp
+
+```shell
+$ hypper search repo -a 'hypper.cattle.io/release-name=.*crd.*' --regexp
+NAME                            	CHART VERSION	APP VERSION	DESCRIPTION                                 
+hypper/fleet-crd                	0.3.500      	0.3.5      	Fleet Manager CustomResourceDefinitions     
+hypper/longhorn-crd             	1.1.001      	           	Installs the CRDs for longhorn.             
+hypper/rancher-backup-crd       	1.0.400      	1.0.4      	Installs the CRDs for rancher-backup.       
+hypper/rancher-cis-benchmark-crd	1.0.301      	           	Installs the CRDs for rancher-cis-benchmark.
+hypper/rancher-gatekeeper-crd   	3.3.001      	           	Installs the CRDs for rancher-gatekeeper.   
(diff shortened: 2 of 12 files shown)


cmd/hypper/search_test.go
```diff
@@ -72,7 +72,32 @@ func TestSearchRepositoriesCmd(t *testing.T) {
 		name:   "search for 'alpine', expect valid yaml output",
 		cmd:    "search repo alpine --output yaml",
 		golden: "output/search-output-yaml.txt",
-	}}
+	}, {
+		name:   "search for hypper, expect latest version 0.2.0",
+		cmd:    "search repo hypper --output yaml",
+		golden: "output/search-hypper-latest.txt",
+	}, { // I use yaml for the output because its easier to create the golden files, sue me :p
```
viccuad, line 79: :rofl:  :rofl:  :rofl:  top!

docs/user/howto/searchannotations.md
```diff
@@ -0,0 +1,31 @@
...
+hypper/rancher-backup-crd       	1.0.400      	1.0.4      	Installs the CRDs for rancher-backup.       
+hypper/rancher-cis-benchmark-crd	1.0.301      	           	Installs the CRDs for rancher-cis-benchmark.
+hypper/rancher-gatekeeper-crd   	3.3.001      	           	Installs the CRDs for rancher-gatekeeper.   
+hypper/rancher-logging-crd      	3.9.002      	           	Installs the CRDs for rancher-logging.      
+hypper/rancher-monitoring-crd   	9.4.204      	           	Installs the CRDs for rancher-monitoring.   
+hypper/rancher-operator-crd     	0.1.300      	0.1.3      	Rancher Operator CustomResourceDefinitions  
+
+```
+
+Of course the usual search flags also work in conjunction with annotations, so you can search for annotations *and* a specific version witht he `--version` flag for example.
```
viccuad, line 31: `witht he` -> `with the`

(doesn't allow me to do suggestions Oo)

pkg/search/search.go
```diff
@@ -80,7 +82,17 @@ func (o *RepoOptions) Run(logger log.Logger, args []string) error {
 	}
 
 	search.SortScore(res)
-	data, err := o.applyConstraint(res)
+	// TODO(itxaka): What happens if we pass more than one annotation? Should we check for all?
+	// Mainly, is this an AND or an OR? Should we just iterate over all the annotations here and
+	// call applyAnnotationsConstraint with each so it works as an AND?
+	// First we need to check for annotations, otherwise the applyConstraint will only return the highest
+	// chart version for a given release
+	data, err := o.applyAnnotationsConstraint(res)
```
viccuad, line 90: since on line 95 we call `o.applConstraint(data)` with the data that we generate here, I suppose it's an AND?
I trust the tests :man_shrugging: .

viccuad (dismissed):
nice! looks nice and clean, the docs are slowly growing :D.

pkg/search/search.go
Itxaka: Goddamm, maybe this is not as explicit as I wanted to be. I know its an AND **currently**! But in the comment Im wondering about what we should do with this, should we treat as AND or maybe as OR? Should we only allow passing one search annotation? etc...

I will update the comment to make it clearer once the search PR is merged and I need to rebase this

Itxaka:
Rebased!

pkg/search/search.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-/* Package search implements the search for charts in repos but extracts it into a package
+/* Package searchimplements the search for charts in repos but extracts it into a package
```
viccuad, line 17:
```suggestion
/* Package search implements the search for charts in repos but extracts it into a package
```

Itxaka: will fix :clown_face:

pkg/search/search.go
```diff
@@ -78,7 +75,17 @@ func (o *RepoOptions) Run(logger log.Logger, args []string) error {
 	}
 
 	SortScore(res)
-	data, err := o.applyConstraint(res)
+	// TODO(itxaka): What happens if we pass more than one annotation? Should we check for all?
```
dragonchaser, line 78: we should either resolve that TODO or remove it from the PR.

pkg/search/index.go
```diff
@@ -14,12 +14,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-/*Package search provides client-side repository searching.
```
dragonchaser, line 17: why did you remove the package description?

dragonchaser (changes requested):

pkg/search/index.go
Itxaka: Now it lives on the doc.go

When there is a bit of text around in the packages, I prefer to extract it to a doc.go file, seems nicer and all the description adn explanation are together.

pkg/search/search.go
Itxaka: still struggling with this, will probably test limiting to just one `-a` flag

mattfarina (changes requested):
I was looking over #93 (the issue with the requirements for this). The idea is that Rancher has 1 charts repo and multiple versions of Rancher use it. Some chart versions will not work with some versions of Rancher. So, a version range of supported rancher versions on the chart could be used to filter for known supported versions.

To make this work we need to have more than a check on the existence of an annotation key/value pair.

For that same issue we need to expose this via a Go API. That can be a separate PR, though.

Itxaka:
I dont understand @mattfarina 

This covers your user case correctly. You can use this to search for `-a "catalog.cattle.io/rancher-version=2.5.*" --regex ` for example and it would search for that with regex in versioning.

So this perfectly covers that.

Now, please, if you have a specific case in mind, **post it in the issue** because if the issue is not clear or not defined enough, we are wasting time writing code that has to be refactored.

This covers much more ground that the specific issue and its flexible enough to be used perfectly on hypper, plus brings value into hypper by making it all about annotations.

viccuad:
What I understood from Matt is that, in addition to filtering on annotations, we want to have a semver-aware filter. So we could ask for things matching `~1.2.3`, or `^1.2.3`, etc.

Itxaka:
> What I understood from Matt is that, in addition to filtering on annotations, we want to have a semver ranges aware filter. So we could ask for things matching ~1.2.3, or ^1.2.3, etc.

but then we are not being consequent here.

Either we want to use a **exact** key to search in the annotations and make that semver aware, or we want this to not be rancher agnostic.

Like if we are gonna search for `catalog.cattle.io/rancher-version` we are tying up ourselves to rancher, but its clear on the ticket that this is not desired....so...Im not sure.

If there is a key we want to search for that we know contains semver then we should say that in the ticket, otherwise we get a generic, not-tied-to-rancher, annotation filter search.

Sure, we can be reckless and just try to load the values from the annotations into a semver constraint and fail otherwise but that is such a brute force approach that it makes no sense to me.


Again, the ticket doesnt seem to be clear. Annotations can be of any type. If we want to search for just one annotation that contains the rancher version and do a semver constraint on there, good for me, I just need to know which annotation so we can discard the rest. Otherwise, lets build a more intelligent, generic annotation search that can be used to search also for **hypper** annotations!


We seem to have an identity issue in which we either want hypper to be this community tool with a clear goal and suddenly we move as making hypper a vessel for rancher use cases. And this makes the hypper vision not clear enough IMO.


In any case, I still think this bring value to hypper as it allows use to center ourselves on annotations and search for those.

pkg/search/search.go
dragonchaser: Can you add a card to the board for that?

Itxaka: This is outdated, doesnt exists anymore. Not sure why github doesnt automatically resolve this....

dragonchaser: cool, will resolve then :)

viccuad (approved):
Approving, the PR is nice and hast tests! If we need to change or add filters, we can do it in subsequent PRs.

dragonchaser (approved):

mattfarina (commented):
Let's hold off on this until after the v0.1.0 release.

In our meeting with Darren he had wanted semver range resolution as a feature. It was more than exact key/value pair checking. I will message him to double check what they are looking for.

mattfarina (approved):

PR #118, opened 2021-04-09
Fix helm settings for flags

Itxaka:
We were filling all the helm settings before parsing the
flags, which means that anything using those helm settings for
any action would just get the default values instead of any cli
values.

This fixes the issue by creating a fill function and calling
it after the flags have been parsed, thus making sure that those
flags are properly propagated.

Otherwise the envsettings behavior its the same as it was before
as the default hypoper values are propagated to helm settings on
Envsettings construct

### DIFF
--- a/pkg/cli/environment.go
+++ b/pkg/cli/environment.go
@@ -106,7 +106,10 @@ func New() *EnvSettings {
 		NoColors: false,
 		NoEmojis: false,
 	}
-	os.Setenv("HELM_NAMESPACE", env.namespace)
+
+	os.Setenv("HELM_NAMESPACE", env.namespace) // WHY???
+
+	// Create default helm settings and propagate our default hypper values
 	env.EnvSettings = cli.New()
 
 	env.EnvSettings.MaxHistory = env.MaxHistory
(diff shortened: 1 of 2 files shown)


pkg/cli/environment.go
```diff
@@ -106,7 +106,10 @@ func New() *EnvSettings {
 		NoColors: false,
 		NoEmojis: false,
 	}
-	os.Setenv("HELM_NAMESPACE", env.namespace)
+
+	os.Setenv("HELM_NAMESPACE", env.namespace) // WHY???
```
viccuad, line 110: @dragonchaser do you remember why?  I don't :smile:.

pkg/cli/environment.go
```diff
@@ -141,6 +144,22 @@ func New() *EnvSettings {
 	return env
 }
 
+// FillHelmSettings will fill the helm cli.Settings with the proper values from CLI flags
+// This has to be called after the flags are set, otherwise the flags are not propagated to Helm cli.Settings
+func (s *EnvSettings) FillHelmSettings() {
```
viccuad, line 149: we could use this function to substitute the assignments in `New()` now!
https://github.com/rancher-sandbox/hypper/pull/118/files#diff-ab0b247f8d73a6afc9c64237096ddb96ca5df0558fe5f5da64f493c2abad4be5R115-R125

viccuad: if we need to call this after the flags are set, I wonder if we could call FillHelmSettings() at the end of AddFlags(), and call it a day.

Itxaka: We cant :cry: I already tried.

We need to wait until the flags are parsed, which happens later than that :_/

viccuad: oh! Thanks for the explanation, resolving :).

viccuad: wait. We could still using the new function in `New()`, for the assignments there, no?

Itxaka: Good point! yes we can!

viccuad (approved):

PR #119, opened 2021-04-13
Add docs generation to workflows

Itxaka:
This uses mdBook to generate a book with all the
docs based of the SUMMARY.md file and with the book.toml
configuration.

Then it builds the book with the docs unders docs in the book/
directory and pushes that to the gh-pages branch which then is
published by github pages, making the docs available


viccuad (approved):
Wooho! Nice!

kkaempf (commented):
Ship it !

PR #125, opened 2021-04-14
feat: Run mdbook test in docs-book gha workflow

viccuad:

### DIFF
--- a/.github/workflows/docs-book.yml
+++ b/.github/workflows/docs-book.yml
@@ -7,17 +7,30 @@ on:
 
 jobs:
   deploy:
+    name: Build, test, and deploy docs
     runs-on: ubuntu-latest
+    env:
+      MDBOOK_VERSION: 0.4.3
+      MDBOOK_LINKCHECK_VERSION: 0.7.0
     steps:
       - uses: actions/checkout@v2
-
-      - name: Setup mdBook
-        uses: peaceiris/actions-mdbook@v1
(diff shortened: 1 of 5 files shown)


.github/workflows/docs-book.yml
```diff
@@ -15,10 +15,12 @@ jobs:
         uses: peaceiris/actions-mdbook@v1
         with:
           mdbook-version: '0.4.3'
+
       - run: mdbook build
+      - run: mdbook test
```
Itxaka, line 20: does this actually works as expected? Running it locally it fails with a spectacular number of failures :o

viccuad: you got me :>. I'm trying to run it locally though, and failing spectacularly..

viccuad: also, for this to work we need to have [mdbook-linkcheck](https://github.com/Michael-F-Bryan/mdbook-linkcheck) and say so in the book.toml.

viccuad:
Tried with the approach of downloading mdbook, mdbook-linkcheck binaries (and caching them).
See results here: https://github.com/viccuad/hypper/runs/2342661504

viccuad:
tested the relative links with `mdbook serve`.

viccuad:
Rebased to actually use gha cache. Example: https://github.com/viccuad/hypper/runs/2342910590?check_suite_focus=true#step:3:11

Itxaka (approved):

PR #142, opened 2021-04-20
fix: Honour annotations' namespaces of shared-deps

viccuad:
Correctly install shared-deps in their own specified namespaces.

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -217,7 +217,7 @@ func (i *Install) InstallSharedDep(dep *chart.Dependency, settings *cli.EnvSetti
 
 	// Set Namespace, Releasename for the install client without reevaluating them
 	// from the dependent:
-	SetNamespace(clientInstall, chartRequested, i.Namespace, true)
+	SetNamespace(clientInstall, chartRequested, i.Namespace, false)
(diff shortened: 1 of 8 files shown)


pkg/action/install.go
```diff
@@ -217,7 +217,7 @@ func (i *Install) InstallSharedDep(dep *chart.Dependency, settings *cli.EnvSetti
 
 	// Set Namespace, Releasename for the install client without reevaluating them
 	// from the dependent:
-	SetNamespace(clientInstall, chartRequested, i.Namespace, true)
+	SetNamespace(clientInstall, chartRequested, i.Namespace, false)
```
Itxaka, line 220: doesnt this needs to be `settings.NamespaceFromFlag` so we respect the flags?

viccuad: It doesn't need to be. The thing is:
- for the dependent chart, we respect the flag, and that gets evaluated at https://github.com/rancher-sandbox/hypper/blob/4bbb1231e2dad98d84f9ac59178a08bc5eb7264e/cmd/hypper/install.go#L121
- for the shared dependencies, we don't respect the flag, as we want them installed on the ns specified in their annotations (we don't want to pass a -n foo, and suddenly install all charts in foo).

mattfarina (approved):

PR #149, opened 2021-04-22
Check for field differences in our settings

Itxaka:
Remove helm settings as composite to avoid propagation of
new fields into our settings automatically
Doesnt really matter as we are overwritting all fields
with our own values
Adds a checker to check that we are not missing any fields

### DIFF
--- a/pkg/cli/environment.go
+++ b/pkg/cli/environment.go
@@ -179,7 +179,7 @@ func (s *EnvSettings) AddFlags(fs *pflag.FlagSet) {
 func (s *EnvSettings) checkCorrectHelmSettings() error {
 	helmSettings := reflect.TypeOf(cli.EnvSettings{})
 	helmNumFields := helmSettings.NumField()
-	hypperSettings := reflect.TypeOf(EnvSettings{EnvSettings: nil})
+	hypperSettings := reflect.TypeOf(EnvSettings{})
 	for i := 0; i < helmNumFields; i++ {
 		field := helmSettings.Field(i)
 		_, found := hypperSettings.FieldByName(field.Name)
--- a/pkg/cli/environment.go
+++ b/pkg/cli/environment.go
@@ -25,7 +25,6 @@ package cli
 import (
 	"fmt"
 	"os"
-	"reflect"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


pkg/cli/environment.go
```diff
@@ -87,8 +86,6 @@ type EnvSettings struct {
 func New() *EnvSettings {
 	// TODO check with reflection that we are not missing any field
 	env := &EnvSettings{
-		EnvSettings: nil, // this is nil it gets overwritten below to ensure that it does not touch Helm's stuff
```
dragonchaser, line 90: If you remove this, this will definitely break things because then the `New()` function of helms cli will definitely be run.

pkg/cli/environment.go
```diff
@@ -107,6 +104,10 @@ func New() *EnvSettings {
 		NoEmojis: false,
 	}
 
+	if err := env.checkCorrectHelmSettings(); err != nil {
+		log.Fatal(err)
+		os.Exit(1)
+	}
 	os.Setenv("HELM_NAMESPACE", env.namespace) // WHY???
```
dragonchaser, line 111: Out of context to this PR, but this is prevents helm code being run on helms data.

pkg/cli/environment.go
Itxaka: do we care? I dont think this is necessary any longer as we are already doing `env.EnvSettings = cli.New()` on line 114 and were already. This line wasn't needed before I think

pkg/cli/environment.go
```diff
@@ -41,11 +43,8 @@ const defaultMaxHistory = 10
 // EnvSettings is a composite type of helm.pkg.env.EnvSettings.
 // describes all of the environment settings.
 type EnvSettings struct {
+	EnvSettings *cli.EnvSettings
```
dragonchaser, line 46: I am not sure if this will still be handled as composition or just as child element. As far as I remember the second will happen....

Itxaka: I think so, the second, because the fields are not propagated into our settings struct anymore

pkg/cli/environment.go
dragonchaser: If I remember correctly there was some weird behaviour with composition, that's why we explicitly set that.

pkg/cli/environment.go
dragonchaser: From that point of view this has become correct yes :)

pkg/cli/environment.go
viccuad: tested it locally, and I think this is line is not needed, indeed.

viccuad (approved):
Nice! :) approving. Tested it also locally, all seems fine.

pkg/cli/environment.go
dragonchaser: then go ahead :)

Issue #90, opened 2021-03-15
Shared Dependency Version Handling


mattfarina:
Shared dependencies (#30) and versions can be problematic. All of shared dependencies and their versions need to resolve across the cluster.

Some thoughts:
- We need to assume the user has access to all relevant namespaces
- We need to assume the user can install/upgrade CRDs (we are already making this assumption)
- It would be nice to have this work out in a performant manner (consider all the secrets that we will need to look through if there are many Helm charts installed)
- How we work out the version will be important. We should suggest a syntax and then look for it on lint. For example, do we want to have a range specified or an individual version and make assumptions? I would suggest we push for ranges (e.g., `^1.2.3`)

kkaempf:
@mattfarina where can I find more information about the caret (`^1.2.3`) syntax ?


Here's [another possible approach](https://guides.rubygems.org/patterns/#declaring-dependencies) on declaring dependencies, esp. the [twiddle-wakka](https://thoughtbot.com/blog/rubys-pessimistic-operator) is useful :wink:

viccuad:
This means that we would not to expand the Release type to save the installed chart version, correct? (I can't find any other way).

viccuad:
@kkaempf I think the semver ranges are this ones https://devhints.io/semver (Helm uses those).

kkaempf:
Thanks, the https://devhints.io/semver link is really helpful !

PR #154, opened 2021-04-30
feat: Add semver range support to all cmd & pkg

viccuad:
Use `"github.com/Masterminds/semver/v3"` to validate, honor, and use the semver constraints listed in shared deps everywhere.
Which right now means:
- `hypper lint CHART`: semver constraint is now a must (we will ERROR if not present).
- `hyper shared-dep list CHART`:  filter against the constraint correctly, now outputting `status: out-of-range` if the installed dep doesn't satisfy what is needed.
- `hypper install CHART`: pull correct chart that satisfies the needed semver constraint. Error if the repository/cache doesn't contain a version that can satisfy it.

In addition, 
- Add related tests. For that, add new `pkg/chart/{Mock(),MockChartOptions()}` functions, since tests for dependencies need releases with charts that contain specific `Chart.Metadata.Version`. I suppose these functions will grow in the future, and we should think about upstreaming them.
- Fix a bug with ns when `hypper shared-dep list -n foo`.

Closes: https://github.com/rancher-sandbox/hypper/issues/90

### DIFF
--- a/cmd/hypper/install.go
+++ b/cmd/hypper/install.go
@@ -65,7 +65,7 @@ func newInstallCmd(actionConfig *action.Configuration, logger log.Logger) *cobra
 			if err != nil {
 				return err
 			}
-			logger.Info(eyecandy.ESPrint(settings.NoEmojis, ":clapping_hands: Done!"))
+			logger.Info(eyecandy.ESPrint(settings.NoEmojis, ":clapping_hands:Done!"))
 			return nil
 		},
 	}
--- a/cmd/hypper/testdata/output/install-with-shared-deps-out-of-range.txt
+++ b/cmd/hypper/testdata/output/install-with-shared-deps-out-of-range.txt
@@ -1,0 +1,3 @@
+🛳  Installing shared dependencies for chart "empty":
+ERROR: ❌  Satisfiable version for chart "testdata/testcharts/shared-dep" not found, aborting
+Error: Satisfiable chart version not found; 0.1.0 is less than 0.3.0
--- a/cmd/hypper/dependency_shared_test.go
+++ b/cmd/hypper/dependency_shared_test.go
@@ -19,6 +19,7 @@ import (
 	"runtime"
 	"testing"
 
+	"github.com/rancher-sandbox/hypper/pkg/chart"
 	"helm.sh/helm/v3/pkg/release"
 )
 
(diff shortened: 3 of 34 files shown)


viccuad:
I haven't updated docs/ with the slightly different ouputs.
(also, it may be a good idea to review the PR stepping through commits, instead of by files).

cmd/hypper/testdata/output/install-with-shared-deps-out-of-range.txt
```diff
@@ -0,0 +1,3 @@
+🛳  Installing shared dependencies for chart "empty":
+ERROR: ❌  Satisfiable version for chart "testdata/testcharts/shared-dep" not found, aborting
+Error: Satisfiable chart version not found; 0.1.0 is less than 0.3.0
```
viccuad, line 3: I don't like this output at all;
I thought that line 2 should just be red (it is) but not contain the `ERROR` in the beginning.
Also, line 3 is a legit error that has been floated upwards, I suppose we are forgetting to print it through our logger. I suppose we can fix it in a following PR.

cmd/hypper/install.go
```diff
@@ -65,7 +65,7 @@ func newInstallCmd(actionConfig *action.Configuration, logger log.Logger) *cobra
 			if err != nil {
 				return err
 			}
-			logger.Info(eyecandy.ESPrint(settings.NoEmojis, ":clapping_hands: Done!"))
+			logger.Info(eyecandy.ESPrint(settings.NoEmojis, ":clapping_hands:Done!"))
```
dragonchaser, line 68: Why did you remove the space? Looks better with....

cmd/hypper/testdata/output/install-with-shared-deps-out-of-range.txt
dragonchaser: You are right, we should add a cart for that (e.g. expand the style guide to include colouring of output).

dragonchaser (approved):

cmd/hypper/install.go
viccuad: it's unclear from the code, but with this change, it looks aligned in my terminal:

tty

Curiously, it looks also aligned on emacs:

emacs

But it doesn't look aligned in vscode:

vscode

nor if I copy-paste the testdata/output and format it here:

```
🛳  Installing shared dependencies for chart "empty":
🛳  - Installing chart "shared-dep-empty" as "my-shared-dep" in namespace "my-shared-dep-ns"…
🛳  Installing chart "empty" as "my-hypper-name" in namespace "hypper"…
👏 Done!
```

I think the clapping-hands emoji is 2 characters wide, and in some places it gets shown in one way, and in others, the other way. What a mess! :man_shrugging: 

I wonder how the ouput from this PR looks in other terminals besides mine! here in xfce and gnome term, it looks ok with the PR change xD.

dragonchaser: :shrug: not a blogger for me, just wanted to point that out, I guess output largely depends on the used terminal....

cmd/hypper/testdata/output/install-with-shared-deps-out-of-range.txt
viccuad: opened https://github.com/rancher-sandbox/hypper/issues/155 !

viccuad:
merging! thanks :)

Issue #55, opened 2021-03-01
optional dependencies


mattfarina:
optional dependencies provide a means to specify dependencies that are not required for installation. For example, there is a chart that install CRDs but not controllers. This CRD chart or the parent can have an optional dependency on the chart that installs the controller. When the parent chart is installed hypper will prompt to ask about installing each optional dependency. There should also be a flag that specifies yes to installing all optional dependencies.

These are similar to suggested dependencies in some systems.

More detail on this design will be added via a design document.

PR #157, opened 2021-05-06
Add optional shared dependencies to hypper

dragonchaser:
This PR adds functionality to add optional shared dependencies to hypper charts through the `hypper.cattle.io/optional-dependencies` annotations. This annotation uses the same syntax as `hypper.cattle.io/shared-dependency`.

This happens in:
- `hypper install CHART [--optional-deps=ask,all,none]`. Selecting `--optional-deps ask` (default) features the ability to install through user interaction (prompt).
- `hypper lint` shows warning on missing `optional-dependencies` annot blocks, and errors if their yaml is misconstructed.
- `hypper shared-dep list` features a new column `type` which contains information about the dependencies. 

In addition:
- Several bugs with namespaces have been fixed.
- Added pkg/cli/environment.go.NewWithNamespace(), used for tests.


go.mod
```diff
@@ -31,3 +32,6 @@ require (
 	k8s.io/client-go v0.20.4
 	sigs.k8s.io/yaml v1.2.0
 )
+
+// manifoldcos code is not thread safe
+replace github.com/manifoldco/promptui => github.com/1lann/promptui v0.8.1-0.20201231190244-d8f2159af2b2
```
viccuad, line 37: We are replacing github.com/manifoldco/promptui with a fork that contains a fix for thread safety (which is only required to run the tests).

Of course, this shouldn't be merged as it is, and we were thinking on either forking into rancher-sandbox, or skipping testing (we would prefer not to).

viccuad:
Closes https://github.com/rancher-sandbox/hypper/issues/55.

go.mod
viccuad: I vote for substituting the library with the builtin reader. I could push this commit to the branch of this PR, if that's ok:
https://github.com/viccuad/hypper/commit/7fe8b3c7fbf41e877167d00dcbc0a7d5ff4ba41d

pkg/action/install.go
```diff
@@ -152,22 +157,46 @@ func (i *Install) InstallAllSharedDeps(chrt *chart.Chart, settings *cli.EnvSetti
...
+				continue
+			}
+		case "ask":
+			if dep.IsOptional {
+				if !promptUserForOptionalDependencyInstall(dep, logger, os.Stdin) {
+					continue
+				}
+			}
+		default:
+			return errors.New("Incorrect value for --optional-deps. Valid values: [default=ask|all|none]")
```
viccuad, line 175: Error output with flags shouldn't be here. We should return an error fit for the library, and catch it on cmd.

pkg/chart/chart.go
```diff
@@ -0,0 +1,68 @@
...
+// Dependency is a composite type of Helm's chart.Dependency
+type Dependency struct {
+	*helmChart.Dependency
+	// hypper specific
+	IsOptional bool `json:"-"`
+}
+
+// GetSharedDeps returns a *[] of all shared and optional dependencies in a
+// chart, read from annotations.
+func GetSharedDeps(c *helmChart.Chart, logger log.Logger) ([]*Dependency, error) {
```
viccuad, line 31: GetSharedDeps() is missing tests.
It would be nice to use pkg/action/action_test.go mocks for it, but, they aren't exported.

go.mod
viccuad: cherry-picked into https://github.com/rancher-sandbox/hypper/pull/157/commits/b7b4c3e41b04c446e85968fdc84867b1c899e2b2.

viccuad (dismissed):
Approving, but I'm 50% approving myself :sweat_smile: .
I will wait for others.

pkg/action/install.go
```diff
@@ -152,22 +167,46 @@ func (i *Install) InstallAllSharedDeps(chrt *chart.Chart, settings *cli.EnvSetti
...
+				continue
+			}
+		case OptionalDepsAsk:
+			if dep.IsOptional {
+				reader := bufio.NewReader(os.Stdin)
+				question := eyecandy.ESPrintf(settings.NoEmojis, ":red_question_mark:Install optional shared dependency \"%s\" ?", dep.Name)
+				if !promptBool(question, reader, logger) {
+					continue
+				}
+			}
```
mattfarina, line 185: If an optional dep is already installed, should a user be prompted about it or just told that the optional dep is already installed? The current flow doesn't seem right.

cmd/hypper/install.go
```diff
@@ -32,8 +32,27 @@ import (
 	logio "github.com/Masterminds/log-go/io"
 	"github.com/rancher-sandbox/hypper/pkg/action"
 	"github.com/rancher-sandbox/hypper/pkg/eyecandy"
+	"github.com/thediveo/enumflag"
```
mattfarina, line 35: I didn't know about this package. Nice find.

mattfarina (changes requested):

pkg/lint/rules/annotations.go
```diff
@@ -40,6 +40,7 @@ func Annotations(linter *support.Linter) {
 	linter.RunLinterRule(support.WarningSev, chartFileName, validateChartHypperRelease(chartFile))
 	linter.RunLinterRule(support.WarningSev, chartFileName, validateChartHypperNamespace(chartFile))
 	linter.RunLinterRule(support.WarningSev, chartFileName, validateChartHypperSharedDeps(chartFile))
+	linter.RunLinterRule(support.WarningSev, chartFileName, validateChartHypperOptionalSharedDeps(chartFile))
```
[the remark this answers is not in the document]
viccuad: We did it in https://github.com/rancher-sandbox/hypper/pull/157/commits/641e92c5bddd5444f7487d843a2969ef5e1599a0.

pkg/action/install.go
viccuad: indeed!
We fixed it to asking only when it isn't installed with https://github.com/rancher-sandbox/hypper/pull/157/commits/f5eb6db6adf5c2fe754f69add62d93e4ddb90380.

pkg/lint/rules/annotations.go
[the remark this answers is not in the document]
mattfarina: We we remove the linter rule for these. Many charts will not have dependencies so there should be no recommendation to add them. These are optional annotations.

mattfarina (approved):

pkg/lint/rules/annotations.go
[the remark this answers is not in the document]
viccuad: Moved them to INFO and reworded output of those that are optional, in https://github.com/rancher-sandbox/hypper/pull/157/commits/fd9228b6a2163e63bc834d1ab8d32446551f55de.

mattfarina (approved):

PR #158, opened 2021-05-18
docs: Update cli output and add optional-dep examples

viccuad:
Update docs with current CLI output and optional-deps examples for
`hypper install`, `hypper lint`, and `hypper shared-deps list`.

Comes together with https://github.com/rancher-sandbox/hypper-charts/pull/5. 

### DIFF
--- a/docs/user/tutorials/quickstart.md
+++ b/docs/user/tutorials/quickstart.md
@@ -70,14 +70,20 @@ all users of the cluster. Think of them as typical system OS libraries/services.
 The commands are the same as you have already used:
 
 ```console
-$ hypper install hypper-charts/fleet --create-namespace
-🛳  Installing chart "fleet" as "fleet" in namespace "fleet-system"…
+$ hypper install hypper-charts/our-app --create-namespace
+🛳  Installing shared dependencies for chart "our-app":
+🛳  - Installing chart "fleet" as "fleet" in namespace "fleet-system"…
+❓ Install optional shared dependency "rancher-tracing" ? [Y/n]:
+y
+🛳  - Installing chart "rancher-tracing" as "rancher-tracing" in namespace "istio-system"…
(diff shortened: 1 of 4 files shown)


docs/user/tutorials/quickstart.md
```diff
@@ -71,8 +71,8 @@ The commands are the same as you have already used:
 
 ```console
 $ hypper install hypper-charts/fleet --create-namespace
-Installing chart "fleet" as "fleet" in namespace "fleet-system"…
-Done! 👏
+🛳  Installing chart "fleet" as "fleet" in namespace "fleet-system"…
+👏 Done!
```
dragonchaser, line 75: I think we should use examples in the quickstart that use shared-dependencies, otherwise the users won't really see the advantage of hypper vs. helm.

dragonchaser (changes requested):

docs/user/tutorials/quickstart.md
viccuad: That's a good point! But I wonder if it would not overcomplicate the quickstart; if we install `our-app` with shared-deps, then we need to explain what happens and why. And the same for the uninstall process and why the unused shared-deps don't disappear yet. The idea of quickstart was to show how to use the CLI in general, not to showcase all the features of Hypper.

dragonchaser: I don't think so, you do not need to explain, just use the same example but use a chart that has shared an optional deps so the users sees that this is possible just by looking at the output.

viccuad: Done in https://github.com/rancher-sandbox/hypper/pull/158/commits/2763662cf6ce21f7098fd6a16f46835009b4c661!

dragonchaser (approved):

PR #159, opened 2021-05-18
feat: Add `hypper install --dry-run` flag

viccuad:


cmd/hypper/dependency_shared.go
```diff
@@ -36,12 +36,13 @@ For example, this Chart.yaml declares two shared dependencies:
...
-	- name: prometheus
-	  version: "13.3.1"
-      repository: "https://example.com/charts"
-	- name: postgresql
-	  version: "10.3.11"
-      repository: "https://another.example.com/charts"
+        - name: prometheus
+          version: "13.3.1"
+          repository: "https://example.com/charts"
+      hypper.cattle.io/optional-dependencies: |
```
viccuad, line 42: these used to be tabs, now they are spaces.

cmd/hypper/install.go
```diff
@@ -101,6 +101,7 @@ func addInstallFlags(cmd *cobra.Command, f *pflag.FlagSet, client *action.Instal
 	f.BoolVar(&client.NoSharedDeps, "no-shared-deps", false, "skip installation of shared dependencies")
 	f.Var(enumflag.New(&optionaldepsmode, "option", OptionalDepsModeIds, enumflag.EnumCaseInsensitive),
 		"optional-deps", "install optional shared dependencies [ask|all|none]")
+	f.BoolVar(&client.DryRun, "dry-run", false, "simulate an install")
```
dragonchaser, line 104: Wouldn't it be nicer to have this as a global flag? I think this would also be helpful for removing stuff and updating namespaces.

cmd/hypper/testdata/output/install-dry-run-with-all-optional-deps.txt
```diff
@@ -0,0 +1,5 @@
+🛳  Installing shared dependencies for chart "empty":
```
dragonchaser, line 1: There should be some output stating that this is a dry run IMHO.

dragonchaser (changes requested):

cmd/hypper/testdata/output/install-dry-run-with-all-optional-deps.txt
viccuad: There is none in Helm, so I went with that.

cmd/hypper/install.go
viccuad: There are some commands (`ls`, `lint`, `shared-dep list`, `repo list`, etc) that may not need it. And since Helm has the flag per command, I went with that too.

dragonchaser: Ok, reasonable approach.

cmd/hypper/testdata/output/install-dry-run-with-all-optional-deps.txt
dragonchaser: I do not like that there either :)

viccuad: :thinking:. To be honest I prefer my commands to be as less verbose as possible: on success, don't print anything, unless is information extracted from running the operation and not known beforehand.
In this case, the user has already passed `--dry-run`, so they know that they are doing a dry run, no need to remind them (there's no way to be safe if users don't even remember what flags and options are they passing).
Also, people use the `--dry-run`, expecting the same output as normal run, for CI and tests.

viccuad: On Helm, doing `--dry-run --debug` will print debug info informing on it. We get that for free, as we ultimately call Helm's action.Install.
It's the same behaviour from Golang utilities: `go vet`, `go build`, etc don't print any known info from passing flags, unless you make it verbose.

dragonchaser: :+1:

dragonchaser (approved):

PR #160, opened 2021-05-19
fix: Lint optional shared-deps, and show errors

viccuad:
We were not parsing the yaml and semver for optional shared-deps.


mattfarina (approved):
Tested and verified it's working.

PR #171, opened 2021-05-28
docs: Add docs/sat-solver.md with justification

viccuad:
Rendered document: https://github.com/rancher-sandbox/hypper/blob/sat-solver/docs/sat-solver.md.

Rendering Latex in GH is a mess.. Rendered them via svgs using this [hack](https://gist.github.com/a-rodin/fef3f543412d6e1ec5b6cf55bf197d7b). See the markdown comment at the beginning of the encoding section.

### DIFF
--- a/docs/sat-solver.md
+++ b/docs/sat-solver.md
@@ -2,13 +2,14 @@
 
 The problem space is defined by a set of packages that depend on each other in
 various ways. A **package** is comprised of:
-- An implicit **chart**.
-- **URI** (repo + name)
-- **semver ranges** (e.g: `~`, `^`). They get collapsed into a specific semver
-  when deployed.
-- Kubernetes **namespace**. Either where to install it, or where are they
+- A **ReleaseName**, either where to installed it, or where it is already
+  installed.
+- A **semver** version.
+- A Kubernetes **namespace**. Either where to install it, or where is it
   installed, if already installed.
-- Defined relations to other packages: **Depends**, **Optional-Depends**.
+- A **Helm Digest** (hash) of the Hypper **chart**.
+- Implicitly, defined relations to other packages: **Depends**,
+  **Optional-Depends**.
 
 Also, not considered now but could be considered in the future:
 - Relations to other packages: **Provides**, **Conflicts**.
--- a/docs/sat-solver.md
+++ b/docs/sat-solver.md
@@ -308,7 +308,7 @@ Then, the objective function is defined as:
 
 <img src="https://render.githubusercontent.com/render/math?math=f_{4} = min( (W_{1} * f_{1}(P) %2B (W_{2} * f_{2}(P)) %2B (W_{3} * f_{3}(P))">
 
-# Existing solver libraries in Golang that we considered
+# Considered solver libraries
 
 Ideally, we want a MAXSAT/Pseudo-Boolean SAT solver library with ample feature
 support, good docs, good API, a thriving community, verbose output, in Golang,
@@ -373,10 +373,12 @@ See also their [docs](https://luet-lab.github.io/docs/docs/concepts/overview/con
 # literature
 
 PBO problem:
+
 - https://arxiv.org/pdf/1007.1022.pdf (CC BY-ND 3.0, yet contains minor errors)
 - https://ths.rwth-aachen.de/wp-content/uploads/sites/4/teaching/theses/grobelna_bachelor.pdf (PB equations optimization)
 
 General literature:
+
 - https://research.swtch.com/version-sat
 - https://arxiv.org/pdf/2011.07851.pdf
 - http://www.lifl.fr/LION9/presentation_orator/lion9.pdf
--- a/docs/sat-solver.md
+++ b/docs/sat-solver.md
@@ -263,17 +263,16 @@ With semver _(a,b,c)_, where
 <img src="https://render.githubusercontent.com/render/math?math=a,b,c \in \mathbb{N}">,
 <img src="https://render.githubusercontent.com/render/math?math=0 \leq a,b,c \leq 10^{3}">,
 the normalized semantic version
-<img src="https://render.githubusercontent.com/render/math?math=\overline{semver}">
-is calculated by constructing a fractional number such that:
+<img src="https://render.githubusercontent.com/render/math?math=\overline{semver}: (a,b,c) \mapsto \mathbb{N}">,
+is defined as:
 
-- the integer part is equal to <img src="https://render.githubusercontent.com/render/math?math=a * 10^{3} %2B b * 10^{0}">
-- the fractional part is equal to _c_.
+<img src="https://render.githubusercontent.com/render/math?math=\overline{semver} = a * 10^{6} %2B b * 10^{3} %2B c * 10^{0}">
 
 E.g:
 
-- `1.2.1 -> 1000 + 2 + 0,1 = 1002,1`
-- `0.46.3 -> 0,463 = 46,3`
-- `7.10.1 -> 7,101 = 7010,1`
+- `1.2.1 -> 1000000 + 2000 + 1 = 1002001`
+- `0.46.3 -> 0 + 46000 + 3 = 46003`
+- `7.10.1 -> 7000000 + 10000 + 1 = 7010001`
 
 The distance
 <img src="https://render.githubusercontent.com/render/math?math=d((a.b.c),(x.y.z))">
--- a/docs/sat-solver.md
+++ b/docs/sat-solver.md
@@ -174,7 +174,7 @@ Although <img src="https://render.githubusercontent.com/render/math?math=x_{1}">
 installed without <img src="https://render.githubusercontent.com/render/math?math=p_{1}">.
 
 
-### Depends on multiple versions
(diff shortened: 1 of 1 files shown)


docs/sat-solver.md
```diff
@@ -0,0 +1,383 @@
...
+<img src="https://render.githubusercontent.com/render/math?math=x_{1} - p_{1} \geq 0">
+
+Installing <img src="https://render.githubusercontent.com/render/math?math=p_{1}"> implies
+installing <img src="https://render.githubusercontent.com/render/math?math=x_{1}">.
+Although <img src="https://render.githubusercontent.com/render/math?math=x_{1}"> may be
+installed without <img src="https://render.githubusercontent.com/render/math?math=p_{1}">.
+
+
+### Depends on multiple versions
+
```
kkaempf, line 178: Not clear what "multiple versions" means here.

docs/sat-solver.md
```diff
@@ -0,0 +1,383 @@
...
+<img src="https://render.githubusercontent.com/render/math?math=x_{1} %2B x_{2} - p_{1} \geq 0">
+
+
+### Conflicts
+
+If <img src="https://render.githubusercontent.com/render/math?math=y_{2}"> conflicts with
+<img src="https://render.githubusercontent.com/render/math?math=x_{1}">:
+
+<img src="https://render.githubusercontent.com/render/math?math=x_{1} %2B y_{3} \leq 1">
```
kkaempf, line 192: Should be `y_{2}` I guess

docs/sat-solver.md
```diff
@@ -0,0 +1,383 @@
...
+
+
+### Provides
+
```
kkaempf, line 196: Not sure if a `Provides` needs to enter the formula.
If `y` requires `p` and `x` provides `p`, then it's reduced to `y` requires `x`

docs/sat-solver.md
```diff
@@ -0,0 +1,383 @@
...
+If <img src="https://render.githubusercontent.com/render/math?math=x_{1}">
+provides
+<img src="https://render.githubusercontent.com/render/math?math=b_{1}">:
+
+<img src="https://render.githubusercontent.com/render/math?math=x_{1} \geq 1">,
+<img src="https://render.githubusercontent.com/render/math?math=b_{1} \geq 1">
+
+### Optional-Depends
```
kkaempf, line 204: Is `optional depends` a helm term or a hypper term ?

docs/sat-solver.md
```diff
@@ -0,0 +1,383 @@
...
+(hence, a Pseudo-Boolean SAT solver).
+
+## [Gini](https://github.com/IRIFrance/gini)
+
+- 1 big known consumer: github.com/operator-framework/operator-lifecycle-manager.
+- Only pure SAT CNF problems.
+
+## [Luet](https://github.com/mudler/luet)
+
+- GPLv3, which would force us to move to GPLv3 given that Golang is usually
```
kkaempf, line 318: Since we know the author of luet, I'd not consider the license issue a major one. :wink:

kkaempf (changes requested):
One typo in the conflicts formula should be fixed.

docs/sat-solver.md
viccuad: True. Maybe better wording is "Depends, that can be satisfied by multiple versions".

docs/sat-solver.md
viccuad: Maybe the definition of Provides is not clear enough. The idea is to use it to create aliases (e.g: chart `foo` depends on chart `relational-db`, and then we have chart `mysql` that provides `relational-db`, and `postgresql` that provides `relational-db`).
The solver needs to be fed with the Provides relation, which is unknown to it, so it can find `x`, in this case `mysql` or `postgresql`.

dragonchaser: We need some interactive stuff for that in case there are multiple candidates either installed on the cluster (e.g. mysql and postgresql) which provide the same or if there is nothing installed on the cluster and there are multiple options available on the repositories that would satisfy the dependency.

docs/sat-solver.md
viccuad: True, we can reword that :). But notice that Luet is just codifying the problem in clauses, which are fed into gophersat. We will need to do that anyways, with more constraints than them (optional dependencies, semvers, provides, etc).

docs/sat-solver.md
viccuad: A hypper one, none of this is provided by Helm.

docs/sat-solver.md
kkaempf: Interactive should be avoided imho, if other heuristics can be used to take a decision.

docs/sat-solver.md
kkaempf: How about using the usual terms as in RPM [weak dependencies](https://rpm.org/user_doc/dependencies.html) ?

docs/sat-solver.md
viccuad: Indeed! thanks :).

viccuad: fixed in https://github.com/rancher-sandbox/hypper/pull/171/commits/8de17228510efc1f06b80190e5b7dd4d6fa0b0db.

docs/sat-solver.md
dragonchaser: @kkaempf agree, but in some cases you have packages that have the same "value" to the product (e.g. mariadb vs. mysql for wordpress etc.) in that case we have to interact with the user or be very opinionated. We should have a list on how to prioritise in such cases (e.g. if anything matches on the cluster use that [override option is needed here], otherwise use our preference)

docs/sat-solver.md
viccuad: reworded in https://github.com/rancher-sandbox/hypper/pull/171/commits/7b7120ebe2cfb7f08def8d16dc5f10457c1ff2f8.

viccuad: reworded in https://github.com/rancher-sandbox/hypper/pull/171/commits/7b7120ebe2cfb7f08def8d16dc5f10457c1ff2f8.

docs/sat-solver.md
dragonchaser: We coined is as optional dependencies in the past, why change now?

viccuad: (and it's already in the last release)

docs/sat-solver.md
viccuad: Also, we maybe should be more clear in the document that the relations "provides" and "conflicts" are not supported yet by Hypper charts, in the "package" definition at the beginning of the document. I was just listing them to help reason about the whole thing.

docs/sat-solver.md
kkaempf: I'd rather not re-invent the wheel :wink:

docs/sat-solver.md
kkaempf: @dragonchaser thinking about our target group (cluster admins), I agree. They should be able to decide mariadb vs. mysql.
But that doesn't hold for other target groups :wink:

kkaempf (dismissed):
Thanks, all good now.

docs/sat-solver.md
viccuad: reworded in https://github.com/rancher-sandbox/hypper/commit/54ac8ea6ccbc07c0181ad76a46812f70767bab4d.

docs/sat-solver.md
viccuad: (clarified that "Provides, Conflicts" are yet not implemented in Hypper with https://github.com/rancher-sandbox/hypper/commit/3046654dd7d03da38bbfa24ce448e3b77cd5207a).

docs/sat-solver.md
```diff
@@ -0,0 +1,439 @@
...
+Given <img src="https://render.githubusercontent.com/render/math?math=P_{1} \dots P_{n}">
+as the new packages targetted to be installed (either existent or new), the
+objective function will be:
+
+<img src="https://render.githubusercontent.com/render/math?math=f_{2}(P) = min(P_{1} %2B \dots %2B P_{n})">
+
+### Maximizing the freshness of packages
+
+When the objective is to install the most recent versions, independently of what
+is already installed.
```
mattfarina, line 240: Ideally we would keep the existing version running rather than upgrading. This was requested from the Rancher folks who would be one of the users of the project. Possibly upgrading if a flag is provided.

viccuad: Sounds good. Do we want to not bump *any* version, or do we want to upgrade to minor and patch versions?

mattfarina: We _do not_ want to upgrade any version. The reason is that they may have passed custom values when installing it. An upgrade could cause those custom values to be lost on upgrade. It is not safe to do an upgrade.

viccuad: That's consistent with the current implementation.
The docs are listing that possibility (and others), in case that in the future we want to implement an "upgrade all to major/minor/patch" command, or a "check cluster", for example.

mattfarina (approved):

PR #176, opened 2021-07-02
Add SAT Solver and use it on `hypper install`

dragonchaser:
This PR adds `internal/{solver,package}/`, which implements a pseudo-Boolean SAT solver. The solver is used to obtain a result on which charts, shared and optional dependencies are going to be installed on `hypper install foo`.

Example: `hypper install hypper-charts/our-app --optional-deps all --debug`

- [x] BUG: non-deterministic order of pkgs in output (which makes goldenfiles differ): order s.PkgResultSet.
- [x] BUG: Loading a chart from local files makes the resulting wantedPkg not to have dependency relations.
- [x] Add solver.Strategy field, implement upgrades in solver by slightly changing constraints creation.
- [x] Skip installing shared and shared-optional dependencies (but solve for them) if `hypper install foo --no-shared-deps`.
- [x] Clobber the code, there's a bunch of commented uneeded code.
- [x] Add `internal/solver/main.go` as a godocs example.
- [x] Fix missing/broken License headers.
- [x] Prettier git history, at least rename WIP commits (quite messy history from pairing).
- [x] Code docs update, including `solver/docs.go`, etc.
- [x] Rework and reenable `pkg/action/install_test.go`.
- [ ] Add more tests to `internal/solver/` and `internal/package/`.
- [x] Make cmd/hypper/{install_test.go, upgrade_test.go} pass.
- [x] Provide repo mock for tests..
- [x] hypper upgrade has a fork in the code: if the chart is not installed, it will install with the SAT solver. If it is to be upgraded, it j   ust calls Helm's action.Install.Run(). We may leave like that for this PR.
- [x] Performance issues with repositories with a lot of chart versions (E.g: https://charts.bitnami.com/bitnami): concurrently create constraints.
- [x] Print and install charts in specific order: first parent, then dependencies: build graph of resulting packages for that.


viccuad:
force-pushed to fix last WIP (so it compiles) and rebased to squash all the `fixup!` commits.

internal/solver/solver.go
```diff
@@ -0,0 +1,485 @@
...
+	}
+
+	// add constraints for all packages that are the same and differ only in version
+	packageConstrs = s.buildConstraintAtMost1(p)
+	constrs = append(constrs, packageConstrs...)
+
+	return constrs
+}
+
+func (s *Solver) Solve(logger log.Logger) {
```
mattfarina, line 134: Why is the logger passed to this method? Would it be better to attach is to the `Solver` type and then log different method details, too?

pkg/action/install.go
```diff
@@ -83,29 +83,98 @@ func CheckDependencies(ch *helmChart.Chart, reqs []*helmChart.Dependency) error
...
 	}
 
-	prefix := ""
-	if lvl > 0 {
-		prefix = fmt.Sprintf("%*s", lvl*2, "- ")
+	s.Solve(logger)
+
+	installedRels := make([]*release.Release, 0)
+	if s.IsSAT() {
+		for _, p := range s.PkgResultSet.ToInstall {
```
mattfarina, line 161: I know I could figure this out by rereading and working through the code again but I think it's faster to ask as I'm going through this, what order will the packages come out of this in?

The order is important. For example, let's say there is a package A that has a dependency on controller B which has a dependency on CRD chart C (A -> B -> C). The install order needs to be C, B, A.

pkg/action/install_test.go
```diff
@@ -79,62 +58,68 @@ func TestInstallAllSharedDeps(t *testing.T) {
...
-			optionalDeps: OptionalDepsNone,
-		},
+		// {
+		// 	name:      "chart metadata has malformed yaml",
+		// 	chart:     buildChart(withMalformedSharedDeps()),
+		// 	golden:    "output/install-malformed-shared-deps.txt",
+		// 	error:     "yaml: line 2: mapping values are not allowed in this context",
+		// 	wantError: true,
+		// 	numRels:   0,
+		// },
```
mattfarina, line 77: Will all these tests be set to work again?

pkg/action/install_test.go
```diff
@@ -196,131 +182,131 @@ func TestInstallAllSharedDeps(t *testing.T) {
...
+// 				Dependency: &helmChart.Dependency{
+// 					Name:       "testdata/charts/vanilla-helm",
+// 					Version:    "0.1.0",
+// 					Repository: "",
+// 				},
+// 				IsOptional: false,
+// 			},
+// 			wantDryRun: true,
+// 			status:     "pending-install",
+// 		},
```
mattfarina, line 209: Will these tests be setup to work again?

mattfarina (commented):
These are from my first pass at the PR. I'll take another pass while trying it out, next.

pkg/action/install_test.go
dragonchaser: Yes, these are the non-working ones we were talking about today.

pkg/action/install_test.go
dragonchaser: Yes, these are the non-working ones we were talking about today.

pkg/action/install.go
viccuad: Indeed, the order is important, and we aren't following it yet.

The solver returns a map of package fingerprints. Right now we just short them with `s.SortPkgSets()`, after separating them in packages to be installed, removed, and unchanged.

I suppose we should solve and then create the resulting graphs (as we have the dependency relations in the package structs). And then iterate through them, installing whatever it is needed. I don't think this is going to land in this PR, though.

(for CRDs, I would be glad to have pre-depends in the future).

internal/solver/solver.go
viccuad: Doh, indeed. Fixed in https://github.com/rancher-sandbox/hypper/pull/176/commits/a6616a6578932de4e7890e40a45a4efae51efccd.

We should refactor more than that, but I would like to have all the tests passing beforehand.

internal/solver/example.go
```diff
@@ -0,0 +1,67 @@
...
+	"bytes"
+	"fmt"
+
+	"github.com/Masterminds/log-go"
+	logcli "github.com/Masterminds/log-go/impl/cli"
+
+	pkg "github.com/rancher-sandbox/hypper/internal/package"
+)
+
+func ExampleSolve() {
```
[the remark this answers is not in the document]
viccuad: indeed, fixed in https://github.com/rancher-sandbox/hypper/pull/176/commits/7e545698c1fb5636eb848a6605be66eeeb365303.

pkg/action/install_test.go
viccuad: Re-enabled `pkg/action/install_test.go` in https://github.com/rancher-sandbox/hypper/pull/176/commits/40733b95d8447d06a061d96e4ddd6a023d5df51c.

Needed refactors and bug fixes for them to pass: 
https://github.com/rancher-sandbox/hypper/pull/176/commits/f9c672ec8fb662ad80515b1aa2d61a83a2c3b8f3
https://github.com/rancher-sandbox/hypper/pull/176/commits/1d790beb9a9554d32174ac8f5cf87d8ad5f09939
https://github.com/rancher-sandbox/hypper/pull/176/commits/b242bb5500c3e6842be7ce5c671c4a00685406d0
https://github.com/rancher-sandbox/hypper/pull/176/commits/647d5db1baeaa2fcd6e2f4091ed4c656394ee161
https://github.com/rancher-sandbox/hypper/pull/176/commits/b2b9f950d7af303f141893564ca89b26448b4edf
https://github.com/rancher-sandbox/hypper/pull/176/commits/43d8c30966bbcd869ab12536705a48a7ed968015
https://github.com/rancher-sandbox/hypper/pull/176/commits/be0ba44489fe2a2bba00197317a51cb92b9b40a5
https://github.com/rancher-sandbox/hypper/pull/176/commits/4f0ed3e9e11546be49b6cfef7b8fd9c919ad5fde
https://github.com/rancher-sandbox/hypper/pull/176/commits/9d5fd6572af4b7b4a47918ae91e71bbc735805da

pkg/action/install_test.go
viccuad: done, see previous comment.

viccuad:
Still working on this. Missing things:
- Make `cmd/hypper/{install_test.go, upgrade_test.go}` pass
- `hypper upgrade` has a fork in the code: if the chart is not installed, it will install with the SAT solver. If it is to be upgraded, it justs calls Helm's action.Install.Run(). We may leave like that for this PR.
- Print and install charts in specific order: first parent, then dependencies: build graph of resulting packages for that.
- Performance issues with repositories with a lot of chart versions (E.g: https://charts.bitnami.com/bitnami):
  The bottleneck is in adding the charts to the package database, and in creating the constraints for them (see with `hypper install --debug`). We should paralelize those tasks.

mattfarina (dismissed):
First, my hypper version is ...

```
❯ hypper version
version.BuildInfo{Version:"v0.2.0-140-g22c9d18", GitCommit:"22c9d1863d205590676c422de496eb5c35bdfa61", GoVersion:"go1.16.5"}
```

This is the latest commit on the resolver branch.

I see a couple issues...

First, I have a chart whose dependency has a dependency. 1 -> 2 -> 3. The dependency of the dependency is not being installed. The repo with the chart is at https://mattfarina.github.io/demo-hypper-charts/repo (the chart name is demo which relies on demo-b which relies on demo-c).

Screen Shot 2021-07-12 at 2 17 52 PM

Strangely, it prompted me about demo-c being an optional dependency which it is not.

The demo dependencies are listed as...
```yaml
annotations:
  hypper.cattle.io/release-name: demoit
  hypper.cattle.io/namespace: demo
  hypper.cattle.io/shared-dependencies: |
    - name: demo-a
      version: "^0.1.0"
      repository: "https://mattfarina.github.io/demo-hypper-charts/repo"
    - name: demo-b
      version: "^0.1.0"
      repository: "https://mattfarina.github.io/demo-hypper-charts/repo"
```

Second, if you look at the image above and the timestamps you'll notice that `demo` appears to be installed before the dependencies. This is backwards. `demo` (the parent) should be installed last. This is important for bases like CRD charts where the CRD chart needs to be installed first.

viccuad:
@mattfarina thanks for looking into it. The bug comes from chart **demo** and **demo-opt** sharing the same release name, `demoit`. On `pkg.action.BuildWorld()` the created packages from them would get merged into one, giving the chart **demo** the optional deps from **demo-opt**.

Fixed the bug by adding pkg.ChartName to the fingerprint and base fingerprint (effectively making it part of the unique key).

A race condition when generating the constraints appeared, and I didn't see why. I have reverted the commit adding that concurrency.

Changing the order of installation of the packages is yet to be done.

mattfarina:
Found another issue. Here's how I ran into it with my testing...

```console
$ hypper install ./demo
🛳  Installing dependencies of chart "demo":
🛳    Installing chart "demo-a" as "demo-a" in namespace "demo-a"…
🛳    Installing chart "demo-b" as "demo-b" in namespace "demo-b"…
🛳  Installing chart "demo" as "demoit" in namespace "demo"…
👏 Done!

$ hypper uninstall demoit -n demo
🔥  uninstalling demoit
✅  release "demoit" uninstalled

$ hypper install ./demo
ERROR: ❌
```

mattfarina:
Found another issue with resolving versions, I think.

The demo chart in v0.1.0 had been set to look for demo-b at `^0.1.0` which means `> 0.1.0 < 0.2.0`. So, it didn't pick up v0.2.0 that depends on the demo-c chart. I updated the demo chart to a v0.2.0 release with the following `Chart.yaml` file...

```yaml
apiVersion: v2
name: demo
description: A Helm chart for Kubernetes
type: application
version: 0.2.0
appVersion: "1.16.0"
annotations:
  hypper.cattle.io/release-name: demoit
  hypper.cattle.io/namespace: demo
  hypper.cattle.io/shared-dependencies: |
    - name: demo-a
      version: "^0.1.0"
      repository: "https://mattfarina.github.io/demo-hypper-charts/repo"
    - name: demo-b
      version: "^0"
      repository: "https://mattfarina.github.io/demo-hypper-charts/repo"
```

On install it still resolves to demo-b at v0.1.0. I get the same result if I set the version for demo-b to `*`. If I explicitly set the demo-b version to 0.2.0 it works to install that. So, it finds it. It just doesn't select it.

_Note, the repo I'm using for testing is https://mattfarina.github.io/demo-hypper-charts/repo_

_Note, each time I'm testing this stuff I'm starting with a fresh Kubernetes setup. I wipe the cluster prior to attempting install. So, there is nothing previously left from a past install_

mattfarina (changes requested):
I'm noting as request changes due the two issues I use reported as comments.

Also, I was looking at the output when installing dependencies of dependencies (transitive ones)...

```console
❯ hypper install ./demo
🛳  Installing dependencies of chart "demo":
🛳    Installing chart "demo-a" as "demo-a" in namespace "demo-a"…
🛳  Installing dependencies of chart "demo-b":
🛳    Installing chart "demo-c" as "demo-c" in namespace "demo-c"…
🛳    Installing chart "demo-b" as "demo-b" in namespace "demo-b"…
🛳  Installing chart "demo" as "demoit" in namespace "demo"…
👏 Done!
```

The indentation of the `Installing dependencies of chart "demo-b":` and `Installing chart "demo-b" as "demo-b" in namespace "demo-b"…` lines doesn't _feel_ right. Should the line `Installing dependencies of chart "demo-b":` even be present?

viccuad:
> The indentation of the Installing dependencies of chart "demo-b": and Installing chart "demo-b" as "demo-b" in namespace "demo-b"… lines doesn't feel right. Should the line Installing dependencies of chart "demo-b": even be present?

Going with the latest repo, `demo-b` depends on `demo-c`. Hence why the line `Installing dependencies of chart "demo-b":` is present.

If there's no indentation, then one can not easily discern which installation comes in because of a dependency.

We could instead do for example:

1.  ```
    ❯ hypper install ./demo
    🛳  Installing dependencies of chart "demo":
    🛳    Installing chart "demo-a" as "demo-a" in namespace "demo-a"…
    🛳    Installing dependencies of chart "demo-b":
    🛳      Installing chart "demo-c" as "demo-c" in namespace "demo-c"…
    🛳    Installing chart "demo-b" as "demo-b" in namespace "demo-b"…
    🛳  Installing chart "demo" as "demoit" in namespace "demo"…
    👏 Done!
    ```

2.  ```
    ❯ hypper install ./demo
    🛳  Installing chart "demo-a" as "demo-a" in namespace "demo-a"…
    🛳  Installing chart "demo-c" as "demo-c" in namespace "demo-c"…
    🛳  Installing chart "demo-b" as "demo-b" in namespace "demo-b"…
    🛳  Installing chart "demo" as "demoit" in namespace "demo"…
    👏 Done!
    ```

mattfarina:
I have another situation I can't quite figure out. I created the situation where I have demo depend on demo-b, demo-b depend on demo-c, and demo-c depend on demo-e. So, demo -> b -> c -> e. Multiple levels. In this situation I cannot get demo-e to install. I did this mostly on disk using `file://` repos.

```console
❯ hypper install ./demo --debug
DEBUG: Original chart version: ""
DEBUG: CHART PATH: /Users/mfarina/code/mattfarina/demo-hypper-charts/charts/demo
DEBUG: Building package DB…
DEBUG: Dependency "demo-d" not found in repos, loading chart
DEBUG: CHART PATH: /Users/mfarina/code/mattfarina/demo-hypper-charts/charts/demo-d
DEBUG: Asking for each optional deps of package demoit_0.2.0_demo_demo if they should be promoted
DEBUG: Building constraints…
DEBUG: Solving…
DEBUG: Result: SAT
🛳  Installing dependencies of chart "demo":
DEBUG: Installing package: fp: demo-a_0.1.0_demo-a_demo-a Currentstate: 0 DesiredState: 0 PinnedVersion: 0.1.0 ChartName: demo-a
DEBUG: CHART PATH: /Users/mfarina/code/mattfarina/demo-hypper-charts/charts/demo-a
DEBUG: Original chart version: "0.1.0"
🛳    Installing chart "demo-a" as "demo-a" in namespace "demo-a"…
DEBUG: creating 1 resource(s)
DEBUG: creating 3 resource(s)
🛳  Installing dependencies of chart "demo-b":
DEBUG: Installing package: fp: demo-c_0.1.0_demo-c_demo-c Currentstate: 0 DesiredState: 0 PinnedVersion: 0.1.0 ChartName: demo-c
DEBUG: CHART PATH: /Users/mfarina/code/mattfarina/demo-hypper-charts/charts/demo-c
DEBUG: Original chart version: "0.1.0"
🛳    Installing chart "demo-c" as "demo-c" in namespace "demo-c"…
DEBUG: creating 1 resource(s)
DEBUG: creating 3 resource(s)
DEBUG: Installing package: fp: demo-b_0.2.0_demo-b_demo-b Currentstate: 0 DesiredState: 0 PinnedVersion: 0.2.0 ChartName: demo-b
          DepRel: &{demo-c demo-c ^0.1.0 demo-c}
DEBUG: CHART PATH: /Users/mfarina/code/mattfarina/demo-hypper-charts/charts/demo-b
DEBUG: Original chart version: "0.2.0"
🛳    Installing chart "demo-b" as "demo-b" in namespace "demo-b"…
DEBUG: creating 1 resource(s)
DEBUG: creating 3 resource(s)
DEBUG: Installing package: fp: demo-d_0.2.0_demo-d_demo-d Currentstate: 0 DesiredState: 0 PinnedVersion: 0.2.0 ChartName: demo-d
DEBUG: CHART PATH: /Users/mfarina/code/mattfarina/demo-hypper-charts/charts/demo-d
DEBUG: Original chart version: "0.2.0"
🛳    Installing chart "demo-d" as "demo-d" in namespace "demo-d"…
DEBUG: creating 1 resource(s)
DEBUG: creating 3 resource(s)
DEBUG: Installing package: fp: demoit_0.2.0_demo_demo Currentstate: 0 DesiredState: 1 PinnedVersion: 0.2.0 ChartName: demo
          DepRel: &{demo-a demo-a ^0.1.0 demo-a}
          DepRel: &{demo-b demo-b 0.2.0 demo-b}
          DepRel: &{demo-d demo-d 0.2.0 demo-d}
DEBUG: Original chart version: "0.2.0"
🛳  Installing chart "demo" as "demoit" in namespace "demo"…
DEBUG: creating 1 resource(s)
DEBUG: creating 3 resource(s)
👏 Done!
```

demo-c has a Chart.yaml of...

```yaml
apiVersion: v2
name: demo-c
description: A Helm chart for Kubernetes
type: application
version: 0.1.0
appVersion: "1.16.0"
annotations:
  hypper.cattle.io/release-name: demo-c
  hypper.cattle.io/namespace: demo-c
  hypper.cattle.io/shared-dependencies: |
    - name: demo-e
      version: "^0.1.0"
      repository: "file://../demo-e"
```

It doesn't appear to try the dependencies at this level. Am I missing something?

The demo-e Chart.yaml is...

```yaml
apiVersion: v2
name: demo-e
description: A Helm chart for Kubernetes
type: application
version: 0.1.0
appVersion: "1.16.0"
annotations:
  hypper.cattle.io/namespace: demo-e
```

demo-c is being pulled from disk yet it's not picking up the dependency in the annotation.

mattfarina:
@viccuad I would go with display option 2.

viccuad:
Implemented display option 2, and added more output when `hypper install --debug`.

mattfarina:
I'll try to walk through my past few comments on re-testing with the latest code...

The issue with the [error](https://github.com/rancher-sandbox/hypper/pull/176#issuecomment-881357781):

> ```console
> $ hypper install ./demo
> ERROR: ❌
> ```

After clearing my hypper cache I can no longer reproduce this issue. If I experience this again I will collect more debugging information and open a new issue. This is good for now.

I like the new indentation, too.

internal/solver/database.go
```diff
@@ -0,0 +1,178 @@
...
+	fp := p.GetFingerPrint()
+	pInDB, ok := pkgdb.mapFingerprintToPkg[fp]
+	if ok {
+		// package already in DB, merge
+		pkgdb.mapFingerprintToPkg[p.GetFingerPrint()] =
+			mergePkgs(*pInDB, *p)
+	} else {
+		// package not there, add it
+		pkgdb.mapFingerprintToPkg[fp] = p
+	}
```
mattfarina, line 165: While I was using a dev/test workflow I ran into a problem. I have a repo with a chart in it. That chart has dependencies. I wanted to test development of those charts without releasing new versions of the charts to the repo. So, I do things like `hypper install ./foo` where `foo` is the directory of the chart. And, foo has a dependency on bar that I'm also developing on. So, I update foo to point to bar in the repo `file://../bar` so it get it from there instead.

This isn't working because the database has foo and bar registered from the charts repo. The fingerprint doesn't take the repo location into account and it appears the instances are merged rather than being separate as they are in different repos.

The instances of charts running in a cluster should not take repos into account. But, when they aren't already running, the chart used needs to be from the right repo location.

mattfarina (changes requested):

internal/solver/database.go
viccuad: I have made a series of commits to tackle this successfully.

**First**, I have implemented it so `file://` repositories are relative to the cwd from where Hypper is being executed. This corresponds with commits https://github.com/rancher-sandbox/hypper/pull/176/commits/91838649cd33a8370b81d764e602572ca5b814b3, https://github.com/rancher-sandbox/hypper/pull/176/commits/e10e3822ed0c7916739c663d4d150783abfe2f7d.

E.g: with chart dependencies of demo listed as following:
```yaml
  hypper.cattle.io/shared-dependencies: |
    - name: demo-a
      version: "^0.1.0"
      repository: "file://charts/demo-a"
```
One does: 
```console
~/local-charts/charts/ $ hypper install ./charts/demo --dry-run
```
where the folder structure is `./charts/{demo, demo-a}`.

**Second**, I have implemented Helm's behaviour where `file://` is relative to the parent chart, in commit https://github.com/rancher-sandbox/hypper/pull/176/commits/6f03e33dc2963f13fa28024c124a8f2985dab6aa. This isn't pretty, as in Helm this is only used when building charts with local dependencies inside, but we need to keep an pass the parent chart absolute path around. I have achieved this by saving the parent chart path into a new Package.ParentChartPath field, with all that entails. 
E.g: with chart dependencies of demo listed as following:
```yaml
  hypper.cattle.io/shared-dependencies: |
    - name: demo-a
      version: "^0.1.0"
      repository: "file://../demo-a"
```
One does: 
```console
~/local-charts/charts/ $ hypper install ./whatever-path-to/demo --dry-run
```
where the folder structure is `./whatever-path-to/{demo, demo-a}`.


Apart from that, added tests, fixed installing when no cache/repositories.yaml is present, and made it work for local nested charts.

mattfarina:
Found one issue that can be dealt with in a follow-up. See #178

mattfarina (approved):

mattfarina:
Approved as things are working in my tests

PR #183, opened 2021-07-29
Fixing db not looking at repos

### LINKED ISSUE #181: Nested Shared Dependency Not Installed
I have demo -> demo-b -> demo-c as a dependency tree. When I go to install demo I see...

```
The following charts are going to be installed:
demo v0.2.0
 ├─ demo-a v0.1.0
 └─ demo-b v0.2.0
```

There is no mention of demo-c and it is not installed

mattfarina:
When f19146a15ac4c750efc2e05ac988b7c04bc598d6 landed the repo
location was switched for the registry location. The registry is
used with OCI. This was likely an accidental autocomplete name
change. It caused the solver to be unaware of the repos and charts
within them. Changing to the right location fixes the issue.

Closes #181


viccuad:
Doh.. thanks for debugging this and finding the culprit :).

viccuad (approved):

PR #184, opened 2021-07-29
Fixing issue where charts in repos not erroring on dep issues

### LINKED ISSUE #182: When required dependency missing install happens without error
I have demo -> demo-b -> demo-c as a dependency tree. When no known location is available for demo-c it silently skips it and completes the installation rather than causing an error. This is a bug.

mattfarina:
The situation is that you have A > B > C. The repo containing B and
C are missing. The code to load files was loading B but not C.
The solver would silently skip C and just install A and B.

For security reasons, we should not install charts from repos
when the repo has not been explicitly added. This is out other
system package managers work and is security guidance out of
reviews of Helm.

The solution here was to look for local files and then throw
an error if a depenency isn't a local file or in a repo.

Closes #182

### DIFF
--- a/pkg/action/world.go
+++ b/pkg/action/world.go
@@ -184,8 +184,8 @@ func (i *Install) CreateDepRelsFromAnnot(p *pkg.Pkg,
 					}
 				}
 				// Local charts (where there is no scheme) and those specified
-				// with the file scheme can be processed. They are local. Any
-				// chart that is in a repository should be from an added repo.
+				// with the file scheme can be processed. They are local. No
+				// dependencies will be pulled from repos not added to Hypper.
 				// Adding a repo enables a user to opt-in for security purposes.
 				// This is what linux system package managers do and it has been
 				// a recommendation from security reviews for Helm.
--- a/pkg/action/world.go
+++ b/pkg/action/world.go
@@ -18,10 +18,13 @@ package action
 
 import (
 	"fmt"
+	"net/url"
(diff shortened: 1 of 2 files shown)


pkg/action/world.go
```diff
@@ -167,40 +171,62 @@ func (i *Install) CreateDepRelsFromAnnot(p *pkg.Pkg,
...
+						// we have an invalid repository
+						return errors.Wrap(err, fmt.Sprintf("unable to parse repository %q", dep.Repository))
+					}
 				}
-				// obtain default ns and release name of dep:
-				depNS = GetNamespaceFromAnnot(depChart.Metadata.Annotations, settings.Namespace())
-				depRelName = GetNameFromAnnot(depChart.Metadata.Annotations, depChart.Name())
+				// Local charts (where there is no scheme) and those specified
+				// with the file scheme can be processed. They are local. Any
+				// chart that is in a repository should be from an added repo.
```
viccuad, line 188: > "Any chart that is in a repository should be from an added repo"
I had a bit of trouble understanding this. We could reword it to "No dependencies will be pulled from repos not added to Hypper" or something like that.

viccuad (dismissed):
Thanks for looking into this and checking for Win paths!

pkg/action/world.go
mattfarina: done

viccuad (approved):

PR #187, opened 2021-08-03
Add shell completion by updating to Cobra 1.2.1

marckhouzam:
I noticed hypper does not yet provide shell completion to its users.
This PR adds it through Cobra 1.2's default completion command.

With this PR:
```
$ bin/hypper completion -h

Generate the autocompletion script for hypper for the specified shell.
See each sub-command's help for details on how to use the generated script.

Usage:
  hypper completion [command]

Available Commands:
  bash        generate the autocompletion script for bash
  fish        generate the autocompletion script for fish
  powershell  generate the autocompletion script for powershell
  zsh         generate the autocompletion script for zsh
```

**Note**: 

If you are not interested in providing shell completion, or if you don't want to use the default Cobra version of the command, it can be disabled (https://github.com/spf13/cobra/blob/master/shell_completions.md#adapting-the-default-completion-command), which would allow to still update Cobra.


mattfarina:
Thanks for the PR @marckhouzam

mattfarina (approved):

PR #190, opened 2021-08-24
Fixing location of repo cache and config in install tests

### LINKED ISSUE #189: make test fails when hypper has repos configured
When you have added repos to hypper (you can see them by running `hypper repo ls`) the tests fail. Hypper tests are using the same environment as hypper on the system. It should be a clean environment apart from the real one for testing.

mattfarina:
Closes #189


dragonchaser (approved):
LGTM

